Issue #1, opened 2016-09-16
Python 3 has no __unicode__() method


millin:
ERRORS:
<class 'django_celery_beat.admin.PeriodicTaskAdmin'>: (admin.E108) The value of 'list_display[0]' refers to '__unicode__', which is not a callable, an attribute of 'PeriodicTaskAdmin', or an attribute or method on 'django_celery_beat.PeriodicTask'.

PR #2, opened 2016-09-16
Python 3 compatibility

millin:
Apply Django build-in decorator python_2_unicode_compatible() for compatibility with Python 2/3 both.

It solves the issue #1 .


millin:
@auvipy, I fixed it.
I mistakenly applied decorator to class that has inherited same decorator from the parent class admin.ModelAdmin. Duplication caused the error for some envs based on PY2.

Issue #8, opened 2016-11-17
Solar schedules


mheppner:
Are there any plans for adding the solar schedules? Can I work on a pull request for this feature?

PR #9, opened 2016-11-18
Add solar schedules

mheppner:
This is in response for #8. I'm not sure what you want the names to be ("solars" for `verbose_name_plural` seems odd, for example). It uses the event names directly from `celery.schedules.solar`, so the model doesn't have any constants. Is it safe to redefine them here?


mheppner:
Migrations fail because the `label` set in `django_celery_beat.apps.BeatConfig` should be `django_celery_beat`, not `beat`.

auvipy:
rebase and try to fix the errors

mheppner:
@auvipy ok this is cleaned up now

django_celery_beat/models.py
```diff
@@ -28,13 +28,61 @@
...
+    )
+    longitude = models.DecimalField(
+        _('latitude'), max_digits=9, decimal_places=6
+    )
+
+    class Meta:
+        """Table information."""
+
+        verbose_name = _('solar')
+        verbose_name_plural = _('solars')
```
ask: maybe this should be 'solar event' and 'solar events'?

mheppner: sounds good to me, thanks ask :)

mheppner: sorry @ask, i forgot to check travis last night. the `unique_together` has been added.

Issue #7, opened 2016-11-15
Django Celery Beat admin updating Cron Schedule Periodic task not taking effect?


kendazheng:
like this 
http://stackoverflow.com/questions/40579804/django-celery-beat-admin-updating-cron-schedule-periodic-task-not-taking-effect

wen96:
Same issue here using `Django` model manager. Whenever I update or create a `PeriodickTask` I have to reload beat to get this changes updated on worker. There a process (I guess pulling process) every 3 minutes on beat saying `Writing entries...` but nothing happens if you don't restart beat. Are we missing something?

Using:
celery 4.0.0
Django 1.9.9
django-celery-beat 1.0.1
Ubuntu 14.04
Redis 2.8.4

Configuration:
```
BROKER_URL = get_setting('BROKER_URL', 'redis://localhost:6379/0')
CELERY_ACCEPT_CONTENT = ['json']
CELERY_TASK_SERIALIZER = 'json'
CELERY_RESULT_SERIALIZER = 'json'
CELERYBEAT_SCHEDULER = 'django_celery_beat.schedulers:DatabaseScheduler'
```

kendazheng:
I have test other celery version and djcelery, it work well. So, it's a bug ?
I have read the source code already. The DatabaseScheduler property method 'schedule' have received the changes, but the method 'reserve' can Not get it. Because I do not know about the project，so I do not know why.

mheppner:
Not sure if this helps, but with Celery 4, your setting name should be prefixed. `BROKER_URL` should be `CELERY_BROKER_URL` and `CELERYBEAT_SCHEDULER` should be `CELERY_BEAT_SCHEDULER`. I don't think you need to specify JSON for any of the serializers because it should be default now.

wen96:
@mheppner yes I know, config was just a copy paste which contains lines which comes from older Celery versions. The worker and beat process shows config on start which is OK and it reads and writes on Redis and on Django database correctly. DB changes simply don't come to `reserve` function as @kendazheng says. Anyway, thank you.

mheppner:
@wen96 that's what I figured, but I thought I'd check anyways. I tried using the django-celery-beat scheduler after I made that comment and I also experienced the same issue. A full restart is needed. I'll look at it tonight and see if I can figure out why.

andrewgleave:
Same here. Any further developments?

mheppner:
[...]

PR #13, opened 2016-12-02
Force to rebuild cached periodic task configuration.

wzywno:
Force to rebuild cached periodic task configuration for Database Scheduler.

Fixes https://github.com/celery/django-celery-beat/issues/7


wen96:
@wzywno why it is closed?

wzywno:
it does not fix every case.

PR #17, opened 2016-12-20
delete CharField null=True

Windsooon:
[Null](https://docs.djangoproject.com/en/1.10/ref/models/fields/#null)

>Avoid using null on string-based fields such as CharField and TextField because empty string values will always be stored as empty strings, not as NULL. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL.

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -244,14 +244,14 @@ class PeriodicTask(models.Model):
         help_text=_('JSON encoded keyword arguments'),
     )
     queue = models.CharField(
-        _('queue'), max_length=200, blank=True, null=True, default=None,
+        _('queue'), max_length=200, blank=True, default=None,
         help_text=_('Queue defined in CELERY_TASK_QUEUES'),
     )
     exchange = models.CharField(
-        _('exchange'), max_length=200, blank=True, null=True, default=None,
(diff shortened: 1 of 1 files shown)


mheppner:
By leaving `default=None`, does Django cast that to an empty string `""`? I don't think a default would be needed. Also, I would think your change needs a migration, right?

vstoykov:
When `null=True` is removed also default should be changes to `''` or omitted. I forgot in which version Django changed default behavior of `CharField` and probably for now will be better to be verbose and to explicitly set `default=''`.

Also Data migration should be made because if there are records in the database with `NULL` values then `ALTER TABLE` will raise an error when trying to set `NOT NULL`.

mheppner:
Not saying it's a bad idea to be more verbose, but I'm pretty sure the default for `CharField` has always been [an empty string](https://github.com/django/django/blob/1.10.5/django/db/models/fields/__init__.py#L773) (`CharField` doesn't override it's inherited `Field` class). Regardless, this shouldn't be merged until a migration is created.

auvipy:
yes anything changes in model should regenerate new migrations

PR #25, opened 2017-02-15
Fix PeriodicTask.enable sync issues

AJRenold:
Hi!

We recently upgraded to Celery 4 and django_celery_beat. We use PeriodTasks and occasionally disable them by setting `PeriodTask.enabled` to False. After upgrading, we noticed that this functionality has stopped working with django_celery_beat.

This pull request adds test cases and fixes for the issues that I observed when troubleshooting the beat command:

1. Calling `Scheduler.reserve(self, entry)` causes the parent class to call `self.schedule[new_entry]` which will add the disabled task back into self._schedule if has recently been run.

2. When the DatabaseScheduler constructor calls `self.update_from_dict(self.app.conf.beat_schedule)`, and entries in beat_schedule will be added to self._schedule, even if the entry has enabled=False in the database.

3. When update is True in `DatabaseScheduler.schedule` and the schedule is fetched by `all_as_schedule`, `self._heap` needs to be set to None so that tick will rebuild the heap which reflects the changed schedule

Happy to discuss these changes! Thanks!

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -249,7 +249,9 @@ class DatabaseScheduler(Scheduler):
         s = {}
         for name, entry_fields in items(mapping):
             try:
-                entry = self.Entry.from_entry(name, app=self.app, **entry_fields)
+                entry = self.Entry.from_entry(name,
+                                              app=self.app,
+                                              **entry_fields)
                 if entry.model.enabled:
                     s[name] = entry
 
@@ -283,6 +285,8 @@ class DatabaseScheduler(Scheduler):
         if update:
             self.sync()
             self._schedule = self.all_as_schedule()
+            # the schedule changed, invalidate the heap in Scheduler.tick
+            self._heap = None
             if logger.isEnabledFor(logging.DEBUG):
                 debug('Current schedule:\n%s', '\n'.join(
                     repr(entry) for entry in values(self._schedule)),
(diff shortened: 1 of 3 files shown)


django_celery_beat/schedulers.py
```diff
@@ -280,6 +285,8 @@ def schedule(self):
         if update:
             self.sync()
             self._schedule = self.all_as_schedule()
+            # the schedule changed, invalidate the heap in Scheduler.tick
+            self._heap = None
```
AJRenold: I'm aware that this is private to the parent class Scheduler and the tick function. Maybe the tick function can have a method to call which indicates if the schedule has been changed since it last ran, if the schedule has changed, then the heap needs to be invalidated because it now has un-sycned task entries

AJRenold:
Also, there appears that this https://github.com/celery/django-celery-beat/pull/9 was merged with a change to models that was not included in migrations

AJRenold:
Ping @ask @auvipy any reviewers available?

auvipy:
try to pass the travis build

AJRenold:
Arg yea... looks like that was the flake8 check on the migration!
```
/home/travis/build/celery/django-celery-beat/django_celery_beat/migrations/0004_auto_20170221_0000.py:17:80: E501 line too long (96 > 79 characters)```

AJRenold:
Thanks!

PR #31, opened 2017-03-09
Django 1.11

vstoykov:


vstoykov:
The error that is raised is very strange. And I'm really confused. This not related to Django 1.11 at all because tests failed with all Django versions.

requirements/test-django111.txt
```diff
@@ -0,0 +1 @@
+django>=1.11b1,<1.12
```
auvipy: this shouls be <2.0

vstoykov: I know that 1.11 should be the last version in 1.x series and next major version planned is 2.0 and not 1.12. Still I think that is better when someone want to restrict package version in `requirements.txt` to `x.y.*` then is better to describe it as `>=x.y,<x.(y+1)` instead of `>=x.y,<(x+1).0` no matter what is the next planned version of that package.

If you still insist that I need to change `1.12` to `2.0` I will change it.

auvipy: we use the suggested way in all of the packages. lets be consistant

vstoykov: Ok I updated my PR.

subha-py:
What happened to this mr?

vstoykov:
I don't know. I've made the changes that @auvipy wanted.

Now I fixed requirements file to point to stable 1.11 and to the beta (because 1.11 is out) and rebased my MR.

There is some problem with travis and test are failing (even in master for change in the documentation). In the MR there is nothing special to add Django 1.11 compatibility. Just test against 1.11, but tests are failing for other versions.

auvipy:
8779c2e46cc4e58aa785f3336753379cc575ac77

PR #32, opened 2017-03-09
Package name should be with dashes not underscores

vstoykov:
There is some problem when `name` is with underscores and not dashes. In some situations `pip` or `wheel` replaces underscores with dashes. When package is installed and in the output of `pip freeze` it is with dashes.

The problem is when asking the XMLRPC API (https://pypi.python.org/pypi) for getting available versions of the package. Then if asked with dashes (as printed by `pip freeze`) then the API does not find the package because it expect to be with underscores.

The easiest way to see the problem is with `manage.py pipchecker` command from `djangno-extensions`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,8 @@ except (AttributeError, ImportError):
     def _pyimp():
         return 'Python'
 
-NAME = 'django_celery_beat'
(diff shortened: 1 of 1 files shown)


vstoykov:
Errors in travis are not related to my PR.

PR #38, opened 2017-04-20
Fix raise ValidationError for simultaneous solar, crontab or interval in PeriodicTask model

sudareva:


django_celery_beat/models.py
```diff
@@ -286,11 +286,14 @@ def validate_unique(self, *args, **kwargs):
...
-        if self.interval and self.crontab and self.solar:
+        err_msg = 'Only one of interval, crontab, or solar must be set'
+        if (self.interval and self.crontab) or (self.crontab and self.solar):
             raise ValidationError({
-                'crontab': [
-                    'Only one of interval, crontab, or solar must be set'
-                ]
+                'crontab': [err_msg]
+            })
+        if self.interval or self.solar:
```
Markez: I think this should be an `and` operator and not `or`. This still raise the ValidationError if only an interval is set or when only solar is set.

PR #40, opened 2017-04-26
Support Periodic Tasks with a start date and one-off tasks

fpghost:
This fork allows PeriodicTasks that only begin after a given date-time. It also allows tasks that only run once before being disabled. 

The 'list_display' of of the PeriodicTask admin was also extended to include the fields: 'interval', 'start_date', 'one_off', which makes verification that tasks have the correct properties much easy in the admin list.


oubeichen:
any update on this?

PR #44, opened 2017-05-09
Add on_delete to ForeignKeys, plus other fixes

### LINKED ISSUE #43: Add `on_delete` to ForeignKeys
When attempting to use celery-beat with the current master Django branch, I get the following error messages:
```
TypeError: __init__() missing 1 required positional argument: 'on_delete'
```
```
django_celery_beat/models.py:176: RemovedInDjango20Warning: on_delete will be a required arg for ForeignKey in Django 2.0. Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior. See https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete
  null=True, blank=True, verbose_name=_('interval'),
```
This is because of the deprecation mentioned here: https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.on_delete
```
Deprecated since version 1.9:
on_delete will become a required argument in Django 2.0. In older versions it defaults to CASCADE.
```

This should be safe to set explicitly to `CASCADE` on all `ForeignKey`s, as this will maintain legacy behavior, and `on_delete` has been an available option since before 1.8, which is the oldest version currently supported by this package.

tarkatronic:
* Fixes #43 -- all ForeignKeys now have an explicit `on_delete=models.CASCADE`. This does not require a migration, as that has been the default.
* This includes the updates from #31, as well as including Django 1.11 in the setup.py
* I also fixed the broken test; it was a problem with comparing naive/aware timestamps.


tarkatronic:
Nope! I checked and this didn't cause any migrations, since the value I
used was previously the default.

auvipy:
cool.

Issue #28, opened 2017-02-22
Add a "Run now" admin action on the scheduled tasks ?


kedare:
Hello,

Would it be interesting to add a "Run now" admin action on the scheduled task? So we could easily run tasks without having to use the shell. This should be quite easy to implement.

Thanks.
Mathieu

auvipy:
a PR is welcome

bradshjg:
@auvipy #52

PlugaruT:
In the admin interface, this action doesn't show up.

bradshjg:
@PlugaruT what are the available actions in the "Action" menu on the "Periodic tasks" page?

PlugaruT:
@bradshjg, only delete action.

bradshjg:
@PlugaruT it looks like the issue is different than missing the "run now" action, you should be seeing the following actions:

* Delete selected periodic tasks
* Enable selected periodic tasks
* Disable selected periodic tasks
* Run selected periodic tasks

PlugaruT:
my bad, I have updated to the latest version and now everything is ok. Thanks!

PR #52, opened 2017-06-10
added run action to periodic tasks admin

bradshjg:
#28 

`current_app.loader.import_default_modules` is sneaky, but it's also sneaky in `TaskSelectWidget` (though it's cached there).


bradshjg:
@auvipy I left the release "date" and "by" section with TODO, let me know if I can help with anything else.

Issue #67, opened 2017-07-28
Interval schedules do not work when Django timezone and Celery timezone are different


DDevine:
As the title says. 
I guess this was an oversight that I should have caught when working on my previous PRs.

marco-silva0000:
I was just finding about this error.
I have the sample tasks running with the default scheduler fine, but when using the database scheduler of this package i have infinite calls on the task, because the is_due call of this type event always is true.
the celery.schedules.schedule class makes this call self.maybe_make_aware(last_run_at)
 print self.maybe_make_aware(last_run_at) gives me this 2017-07-28 10:27:06.436527 when my current timezone is one hour more, so the rem_delta = -1 day, 23:23:08.922251
i have the following settings
TIME_ZONE = 'Europe/Lisbon'
CELERY_ENABLE_UTC = True
CELERY_TIMEZONE = 'Europe/Lisbon'

I can help fixing this if you give me any pointers, in the meanwhile, is there som other version that does not have this problem?

DDevine:
There has been some recent changes to Celery, and I've also fixed some things too - notably I have made last_run_at timezone aware which may solve your issue because the delta might be calculated correctly. 
You could try the code in this pull request: https://github.com/celery/django-celery-beat/pull/68

2ps:
This issue is likely caused by stupidity in Celery 4.1.0 -- where they couldn't figure out how to calculate now properly for non-UTC timezones.  You can fix it by subclassing from `celery.Celery` and writing your own `now` function that returns a properly timezoned now (e.g., `datetime.now(my_timezone)`).

DDevine:
@2ps is that a fix or a hack?

Patches welcome - negativity isn't. Remember, contributors (and/or their employers or customers) pay a price for the work they put in. Put your patches where you mouth is.

harelw:
Looks like most recent Celery takes inherits timezeone from Django by default. I ran into problems when I set CELERY_TIMEZONE to the same value as TIME_ZONE manually. The fix was simply to remove CELERY_TIMEZONE and null out last_run_at for each task.

gdvalderrama:
@harelw Could you detail what problems you ran into?
I was just trying to do that since beat is using UTC instead of TIME_ZONE.

qburst-starsona:
You can try the `nowfun` argument in `crontab` for setting different timezone

https://stackoverflow.com/questions/21827290/celery-beat-different-time-zone-per-task

auvipy:
[...]

Issue #30, opened 2017-03-03
Code for resetting "last run time" not working


gitaarik:
This code in the README:

```
from django_celery_beat import PeriodicTask, PeriodicTasks
PeriodicTask.objects.all().update(last_run_at=None)
PeriodicTasks.changed()
```

I get an import error for the first line, and if I fix that I get a error for the third line saying the `changed()` method requires an argument `instance`.

Should the code be like this?

```
from django_celery_beat.models import PeriodicTask, PeriodicTasks
PeriodicTask.objects.all().update(last_run_at=None)
[PeriodicTasks.changed(task) for task in PeriodicTask.objects.all()]
```

famdude:
so what is the solution?!

PR #68, opened 2017-07-28
Fix interval schedules by providing nowfun.

DDevine:
Fix for https://github.com/celery/django-celery-beat/issues/67

I have thoroughly tested that solar, interval and crontab tasks all work as expected when run simultaneously with my changes.

Also fixing solar test.

Django_celery_beat is approaching the point of actually being usable ;)

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -60,9 +60,9 @@ class SolarSchedule(models.Model):
 
     @property
     def schedule(self):
-        return schedules.solar(self.event, 
-                               self.latitude, 
-                               self.longitude, 
+        return schedules.solar(self.event,
+                               self.latitude,
+                               self.longitude,
                                nowfun=lambda: make_aware(now()))
 
     @classmethod
@@ -112,7 +112,10 @@ class IntervalSchedule(models.Model):
 
     @property
     def schedule(self):
-        return schedules.schedule(timedelta(**{self.period: self.every}))
+        return schedules.schedule(
+            timedelta(**{self.period: self.every}),
+            nowfun=lambda: make_aware(now())
+        )
 
     @classmethod
     def from_schedule(cls, schedule, period=SECONDS):
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -13,6 +13,8 @@ from celery.utils.encoding import safe_str, safe_repr
 from celery.utils.log import get_logger
 from kombu.utils.json import dumps, loads
 
+from datetime import timezone
+
 from django.db import transaction
 from django.db.utils import DatabaseError
 from django.core.exceptions import ObjectDoesNotExist
@@ -85,10 +87,7 @@ class ModelEntry(ScheduleEntry):
 
         if not model.last_run_at:
             model.last_run_at = self._default_now()
-        orig = self.last_run_at = model.last_run_at
-        if not is_naive(self.last_run_at):
-            self.last_run_at = self.last_run_at.replace(tzinfo=None)
-        assert orig.hour == self.last_run_at.hour  # timezone sanity
+        self.last_run_at = make_aware(model.last_run_at)
 
     def _disable(self, model):
         model.no_changes = True
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -13,8 +13,6 @@ from celery.utils.encoding import safe_str, safe_repr
 from celery.utils.log import get_logger
 from kombu.utils.json import dumps, loads
 
-from datetime import timezone
-
 from django.db import transaction
 from django.db.utils import DatabaseError
 from django.core.exceptions import ObjectDoesNotExist
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -98,7 +98,11 @@ class ModelEntry(ScheduleEntry):
         return self.schedule.is_due(self.last_run_at)
 
     def _default_now(self):
-        return self.app.now()
+        now = self.app.now()
+        # The PyTZ datetime must be localised for the Django-Celery-Beat
+        # scheduler to work. Keep in mind that timezone arithmatic
+        # with a localized timezone may be inaccurate.
+        return now.tzinfo.localize(now.replace(tzinfo=None))
 
     def __next__(self):
         self.model.last_run_at = self.app.now()
@@ -113,7 +117,6 @@ class ModelEntry(ScheduleEntry):
         obj = type(self.model)._default_manager.get(pk=self.model.pk)
         for field in self.save_fields:
             setattr(obj, field, getattr(self.model, field))
-        obj.last_run_at = make_aware(obj.last_run_at)
         obj.save()
 
     @classmethod
--- a/t/unit/test_schedulers.py
+++ b/t/unit/test_schedulers.py
@@ -444,7 +444,7 @@ class test_models(SchedulerCase):
 
         assert s.schedule is not None
         isdue, nextcheck = s.schedule.is_due(dt_lastrun)
-        assert isdue is False  # False means task is not due, but keep checking.
+        assert isdue is False  # False means task isn't due, but keep checking.
         assert (nextcheck > 0) and (isdue is False) or \
             (nextcheck == s.max_interval) and (isdue is True)
 
--- a/t/unit/test_schedulers.py
+++ b/t/unit/test_schedulers.py
@@ -439,15 +439,24 @@ class test_models(SchedulerCase):
 
     def test_SolarSchedule_schedule(self):
         s = SolarSchedule(event='solar_noon', latitude=48.06, longitude=12.86)
-        dt = datetime(day=25, month=7, year=2017, hour=12, minute=0)
+        dt = datetime(day=26, month=7, year=2050, hour=1, minute=0)
         dt_lastrun = make_aware(dt)
 
         assert s.schedule is not None
-
         isdue, nextcheck = s.schedule.is_due(dt_lastrun)
-        assert isdue is False  # False means scheduler needs to keep checking.
+        assert isdue is False  # False means task is not due, but keep checking.
         assert (nextcheck > 0) and (isdue is False) or \
             (nextcheck == s.max_interval) and (isdue is True)
+
+        s2 = SolarSchedule(event='solar_noon', latitude=48.06, longitude=12.86)
+        dt2 = datetime(day=26, month=7, year=2000, hour=1, minute=0)
+        dt2_lastrun = make_aware(dt2)
+
+        assert s2.schedule is not None
+        isdue2, nextcheck2 = s2.schedule.is_due(dt2_lastrun)
+        assert isdue2 is True  # True means task is due and should run.
+        assert (nextcheck2 > 0) and (isdue2 is True) or \
+            (nextcheck2 == s2.max_interval) and (isdue2 is False)
 
 
 @pytest.mark.django_db()
--- a/t/unit/test_schedulers.py
+++ b/t/unit/test_schedulers.py
@@ -118,11 +118,11 @@ class test_ModelEntry(SchedulerCase):
         assert e.options['routing_key'] == 'cpu'
 
         right_now = self.app.now()
-        
         m2 = self.create_model_interval(
             schedule(timedelta(seconds=10)),
             last_run_at=right_now,
         )
+
         assert m2.last_run_at
         e2 = self.Entry(m2, app=self.app)
         assert e2.last_run_at is right_now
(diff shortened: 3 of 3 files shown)


DDevine:
Ephem is failing to install in the build environment. I have no idea why...

DDevine:
In regards to https://github.com/celery/celery/pull/4170 this pull-request also fixes broken crontabs that arose from https://github.com/celery/celery/pull/4173.

razerraz:
It seems this commit don't resolve everything :
https://github.com/celery/django-celery-beat/issues/70#issuecomment-320624147

pztrick:
On master, when I am using both Django and Celery timezone as `America/New_York`, this task runs incessantly (#4041) and ignores the interval:

```
CELERYBEAT_SCHEDULE = {
    'debug': {
        'task': 'app.tasks.debug',
        'schedule': 10.0  # every 10 seconds
    }
}
```

Cherry-picking this pull request's commit https://github.com/celery/django-celery-beat/pull/68/commits/2312ab540910a06d7c1b285ed8d09e7576791021 fixes it.

However, for some reason the complete pull request causes the task to only be run time (rather than in 10s interval) -- so not fixed?

DDevine:
Hey @pztrick thanks for looking into this.
Have you tried using the snippet described in https://github.com/celery/django-celery-beat/issues/30 to clear last_run_at? 
My system heavily uses interval schedules - so I don't see this problem.

pztrick:
@DDevine I performed this in shell and relaunched application. It still only executes the task once at startup:

```
from django_celery_beat.models import PeriodicTask, PeriodicTasks
PeriodicTask.objects.all().update(last_run_at=None)
[PeriodicTasks.changed(task) for task in PeriodicTask.objects.all()]
```

thanks!

DDevine:
I don't really know what could be going wrong to be honest.

TIME_ZONE = 'America/New_York' # Django TZ
CELERY_ENABLE_UTC = True # If false, when messages are sent they are kept in local time, if True they are converted to UTC.
CELERY_TIMEZONE = 'America/New_York'

Are these your settings? Though I think CELERY_TIMEZONE can be UTC as well and it worked for me.

pztrick:
@DDevine, I have ENABLE_UTC disabled.

```
TIME_ZONE = 'America/New_York'
CELERY_ENABLE_UTC = False
CELERY_TIMEZONE = 'America/New_York'
```

@marco-silva0000 indicated [here](https://github.com/celery/celery/issues/4041#issuecomment-321107319) that he was going to test the commit earlier. He could also try testing this pull-request also to see if he can reproduce?

marco-silva0000:
I will

DDevine:
This should work without CELERY_ENABLE_UTC - so I guess my patch needs work. However, is enabling CELERY_ENABLE_UTC an option for you?

razerraz:
I've tried further things on my project, taking care of your recommandations - celery from git, beat from your repo.

Project default : 
USE_TZ = True - TIME_ZONE = 'Europe/Paris' - CELERY_TIME_ZONE = TIME_ZONE - CELERY_ENABLE_UTC = True

In this case, nothing work, even with reseting last_run_at and manualy apply changes as you mentionned.

The only way to get things to work is with project settings ;
USE_TZ = False - CELERY_ENABLE_UTC = False - no CELERY_TIME_ZONE set

Furthermore, I have to reset last_run_at and apply changes every time a task is edited in django admin, otherwise it stopped working.

This way, since periodic tasks seems ok, crontab tasks are repeated indefinitly, with no other way than killing beat process and deleting the task.

Is it something wrong with the way celery handle things on schedulers right now ? I'm just looking at this as a naive user waiting for it's needs (database scheduler tasks), the efforts you did for that, I wonder why it seems to be so problematic to get rid of this regression, since all was working fine before celery 4...

Cheers

DDevine:
Yes, Celery 4 and the deprecation of djcelery in favour of django-celery-beat were a major shake-up. My system requires Solar scheduling (in addition to interval and crontab) so as you can imagine I was excited to find that apparently Celery handled this... Surprise!

I will do a fresh install of my environment to try and see if I can replicate your issue. Can you do a `pip freeze` and share versions of celery, django-celery-beat, kombu, billiard, django, python, etc.

One problem I have encountered before is not explicitly uninstalling an old pip dependency before installing a new one - pip just sees the old version and doesn't put the new version over the top even if you're pulling from git because it just sees that the version numbers are the same.

razerraz:
Installation is clean, I  don't use pip but arch packages builds, and the old versions was removed before merging new ones.  The way arch handle things on upgrades leave almost nothing to doubt on that.
The project can be however source of breaking. It's a big one with a bunch of modern stuff like asyncio, django-channels. I'll try to start with empty project.

marco-silva0000:
Razerraz, you can use an example project I created here
https://github.com/marco-silva0000/celery/commit/4c99c06073e862b6d0b91b7b8d48c76587acc75a

razerraz:
Definitly, something goes really wrong with timezones...
I play with django shell, configuration values related to timezones, and I get something so messy I don't have a clue where to start looking...
For instance, setting USE_TZ = True, changing timezone from 'Europe/Paris' to 'America/chicago' with a fresh clean of tasks, crontabs & periods between changes gives a very different behavior.
Living in Europe gives no scheduled tasks working at all, with only to fix last_update_on to datetime.now() on the django shell. This way the task is loaded one time and bye bye.
When, living in chicago, all the tasks are flooded (even celery.backend_cleanup !) several times per second

So it seems that local time of the host changes the way things are handled dramaticly, I wonder if that's expected behaviour...

pztrick:
Just to share my environment...

```
# settings.py

# django timezone
TIME_ZONE = 'America/New_York'
USE_TZ = True

# celery timezone
CELERY_TIMEZONE = TIME_ZONE
CELERY_ENABLE_UTC = False
CELERYBEAT_SCHEDULER = 'django_celery_beat.schedulers:DatabaseScheduler'
```

```
# requirements.txt

# django-celery-beat @ master + commit b6c0581
# "Removing code that forced last_run_at to be timezone naive for no reason"
git+git://github.com/pztrick/django-celery-beat@83630707b900b526da924d030dfd65acbb739d72#egg=django_celery_beat

# celery @ master
git+git://github.com/celery/celery@bd347f7565f3a72c8cfb686ea0bfe38cfa76e09b#egg=celery
```

My code runs flawlessly on these two commits. I can change/update/delete cron schedules at runtime and scheduler changes take impact immediately without having to restart the celery beat daemon. All task datetimes are in the target `America/New_York` timezone.

Running on this P/R however was not working. I do not have time to test today but others can try my `requirements.txt` possibly?

---

Also, one nit I have with the `DatabaseScheduler` is that it picks up `settings.CELERYBEAT_SCHEDULE` periodic tasks (which is good), but if you remove these tasks later from your `CELERYBEAT_SCHEDULE` they are still persisted in the database (which may not be intuitive). I personally would like a Boolean bit on the database model to purge these periodic tasks at initialization if they no longer are defined in the `CELERYBEAT_SCHEDULE` (and that is where they were created). Would a P/R be welcome -- to allow about source code-defined and dynamic/ORM defined periodic tasks?

razerraz:
On a fresh project, it seems to work with your timezone but not mine. I'll do further check to confirm that, but it seems to be this way...

DDevine:
@razerraz Keep in mind that the django-celery-beat documentation specifically says:
"""
The database scheduler won’t reset when timezone related settings change, so you must do this manually 
"""
So if you tested with my timezone and then just changed the timezone in settings.py - this will not work. You must run the code to reset the last_run_at times after you change the timezone in settings.py

razerraz:
I've tried almost everything with a full clean project, changing broker from redis to rabbitmq, database from postgres to sqlite... It just don't work at all...
When I change the timezone I do a full database clean and migration
I've sometime a task working, always the same, with crontab. Other added tasks are never runned, periodic or crontab, and the last_run_at stays null on the database. That's with my timezone
With yours (and after full database clean), tasks are flooded and I have to kill everything.

Indeed, it's host timezone dependent, since it seems to work for you.
But I'm pretty sure something goes wrong with that, try with my timezone, I bet it will not work.

harukaeru:
Same problem is happened on my Mac.
I created a simple project. So check this whether it does not work or works well.

`master` branch uses `django-celery-beat==1.0.1` and `DDevine_timezone_fix` is this PR.

Here is the repository
https://github.com/harukaeru/django-celery-beat-example/tree/DDevine_timezone_fix

pztrick:
I feel like `last_run_at` should only be evaluated/meaningful for Interval tasks. Cron scheduler tasks should be fine resetting to `None` and being re-scheduled every reboot, no?

```
The database scheduler won’t reset when timezone related settings change, so you must do this manually
```

Is there any good reason for this behavior besides "that's just how it's always been"? (Again, understanding `last_run_at` needs to be persisted for interval tasks.)

pztrick:
@harukaeru You also need to use celery@master and not `celery==4.1.0`. 

You should try using this P/R and celery@master. If that doesn't work, you can try my commit which cherry-picks just one commit that I found fixed it for me:

```
# requirements.txt

# django-celery-beat @ master + commit b6c0581
# "Removing code that forced last_run_at to be timezone naive for no reason"
git+git://github.com/pztrick/django-celery-beat@83630707b900b526da924d030dfd65acbb739d72#egg=django_celery_beat

# celery @ master
git+git://github.com/celery/celery@bd347f7565f3a72c8cfb686ea0bfe38cfa76e09b#egg=celery
```

harukaeru:
@pztrick Thank you!

But I've noticed actually my problem is totally different from the problem described here.
I mistakenly read all over issues because I skim-read too much. 
I assume it is almost not related here... :(
So your commit did not solve my problem but thanks for your help, so that I could notice.

After commenting in the above, I firmly read `django-celery-beat` and `celery` sources and then I solved my problem with overriding `schedules_equal` method in `DatabaseScheduler` class.
In this method, it is simply compared by `schedule` (like `CrontabSchedule` class) only.

But this modification of mine may be not related here. It just solves my problem.
I think my problem is not caused by a bug but merely outside the scope of `django-celery-beat`.

I'm sorry to bother your hand :(

DDevine:
This push should fix installations in fresh environments - in my case it was causing tasks to be constantly due.

pztrick:
Hi @DDevine I think this latest commit will fix. I seem to remember seeing a timezone-naive datetime somewhere for `now()` when I was originally debugging that was in-fact the accurate nominal hours for the local timezone but didn't have any tzinfo set... so it would get re-offset a second time and depending on whether you were ahead or behind UTC you end up running the task repeatedly or never.

Will hopefully be able to test tomorrow but if not next week. :) Thanks!

kodeine:
how to update to latest commit using pip?

pztrick:
This P/R now working on my system as expected!

I am using latest celery@master commit as below.

```
# reqs.txt
git+git://github.com/DDevine/django-celery-beat@6c92cc67ca2a2494215d5557f2d5a5f3777da145#egg=django_celery_beat
git+git://github.com/celery/celery@a20e7922b3641d03110df2b770b7480c3f32b42c#egg=celery
```

DDevine:
Does anybody have some guidance on how the PyPy compatibility issue should be dealt with? Feature gate based on whether PyEphem is present?

bofia:
I also had the issue with tasks constantly getting run (1000s per minute).

I've tried the commits noted by @pztrick and it seemed to stop constantly running, but now the tasks are never sent:

The logs just constantly show "beat: Waking up in 5.00 seconds." even though i have a PeriodicTask scheduled for an interval of 30s.

My settings are:

```
TIME_ZONE = 'America/New_York'
CELERY_TIMEZONE = 'America/New_York'
CELERY_ENABLE_UTC = True
```

keeth:
@pztrick's setup above works for me too.

```
TIME_ZONE = 'America/Los_Angeles'
CELERY_TIMEZONE = TIME_ZONE
```

I found that my crontab tasks did not run at all with this PR and celery 4.1, but using celery@master fixed it. 

Best of all django-celery-beat is no longer DOS'ing my server with zillions of copies of the same task.

Wish I could help with the PyPy business, but I'm a Python newb. Thanks for the fix tho!!

xeor:
It seams that there are a couple of issues related to this problem now... It seams like 4.2 might fixing some/all of them..

My django-celery-beat is not doing anything at the moment, just waking up a lot :(

DDevine:
@xeor That is true to some extent - Celery 4.2 is needed in order for this patch to be meaningful as it currently requires Celery from master to work. However, the issue is between Celery and django-celery-beat

pztrick:
My Celery task is firing an hour early now after United States (my timezone) daylight savings time occurred. Will see if it happens again after restarting application. Commenting here as I am working off of this PR branch now.

thedrow:
Sorry this took so long to merge :)
I'll release this fix as soon as I can.

PR #78, opened 2017-08-31
Support Periodic Tasks with a start date and one-off tasks 

oubeichen:
modified from https://github.com/celery/django-celery-beat/pull/40

rebased, conflict fixed and unit tests added


django_celery_beat/migrations/0005_auto_20170425_1819.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
auvipy: this migration file needed to be deleted and regenerated after pulling from master

yywing:
When will this mr be merge?

oubeichen:
@auvipy pull request updated

reverland:
Awesome work， I'm just trying hard to continue work on your PR and see your update. Thanks. @oubeichen

reverland:
@oubeichen the flake8 check failed in CI

oubeichen:
All checks have passed, finally...

django_celery_beat/schedulers.py
```diff
@@ -96,6 +97,22 @@ def _disable(self, model):
     def is_due(self):
         if not self.model.enabled:
             return False, 5.0   # 5 second delay for re-enable.
+
+        # START DATE: only run after the `start_time`, if one exists.
+        if self.model.start_time is not None:
+            if maybe_make_aware(self._default_now()) < self.model.start_time:
```
viiicky: I understand this has already been merged, but I am looking forward to use this feature once the next release goes live and was just wondering if this line would cause any issue, when the `self.model.start_time` is in a different timezone as compared to `self._default_now()`?

cc @oubeichen @auvipy

viiicky:
Well, this has been already merged, but with these changes it would really be helpful that when `start_time` and `one_off` are both set one can omit the `interval`/`schedule` fields while creating  `periodictask`s. This way it can basically be used to fire a "one time" "specific time" task with doing something like this:
- PeriodicTask.objects.create(name='Wish India', task='tasks.wish', args=json.dump(['Namaste']), start_time=datetime.datetime.now(2019, 1, 1, 0, 0, 0, tzinfo=IST), one_off=True)
- PeriodicTask.objects.create(name='Wish Spain', task='tasks.wish', args=json.dump(['Hello']), start_time=datetime.datetime(2019, 1, 1, 0, 0, 0, tzinfo=EST), one_off=True)

viiicky:
And also may be instead of having `one_off` as option, maybe the more flexible option would have been to have a number as input something like `number_of_times`, that can be set to decide how many times the task should run.

Still, waiting for this to get merged as soon as possible 👍

Issue #83, opened 2017-09-15
Related model 'django_celery_beat.CrontabSchedule' cannot be resolved


pashinin:
Was trying to intall from git. 

1. When run `manage.py migrate` I got this:

```
Related model 'django_celery_beat.CrontabSchedule' cannot be resolved
```

2. Also when I did `makemigrations` I got:

```
Permission denied: '../site-packages/django_celery_beat/migrations/0005_auto_20170915_1139.py'
```

There is no such file actually in migrations.

```
- Alter field event on solarschedule
File "/usr/local/lib/python3.6/site-packages/django/core/management/commands/makemigrations.py", line 232, in write_migration_files
    with io.open(writer.path, "w", encoding='utf-8') as fh:
PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.6/site-packages/django_celery_beat/migrations/0005_auto_20170915_1139.py'
```

Well, it wants to write a file, not read it. But still maybe there is unpushed migration in your files or something else?

pashinin:
Fixed 2nd problem here: https://github.com/pashinin/django-celery-beat/commit/a87992e605fd6de026cedda74403d97f1d8ffde4

auvipy:
https://github.com/celery/django-celery-beat/pull/84#issuecomment-345972005

PR #84, opened 2017-09-15
Fix creation of a new migration each "makemigrations"

pashinin:
#83


brianmay:
Note #92  appears to be a fix for the same problem.

syurchenko:
`[(x, _(x)) for x in schedules.solar._all_events].sort()` returns None. This is incorrect. Choices tuple becomes empty.
`[(x, _(x)) for x in sorted(schedules.solar._all_events)]` should be used instead.

pashinin:
Use https://github.com/celery/django-celery-beat/pull/92

PR #86, opened 2017-09-19
Fix CrontabSchedule example

hamishcampbell:
A cron expression of ``30 * * * *`` will (correctly) execute hourly at 30 minutes past the hour.


DDevine:
Yes you are correct. Patch looks good.

PR #94, opened 2017-11-02
remove ephem requirement

ransford:
No other files in this repository mention ephem, so it looks like this rather heavyweight library is a stale dependency that is no longer needed.  Also, since pip seems to want to build a shared library from source when installing ephem on Linux and Darwin, including ephem as a dependency to django_celery_beat has the (presumably) unintended consequence of requiring Python development headers.


dvdmgl:
Also there is an issue [building ephen under pypy3](https://github.com/brandon-rhodes/pyephem/issues/126)

mpiskore:
It would be awesome to merge this PR since this produces install errors both on my local setup and our Jenkins machine and this branch is 9 commits behind master. I can create my own fork and use that but I believe that everyone could benefit from keeping it simple.

ransford:
Now it's zero commits behind master!

DDevine:
Ephem is needed for Solar Scheduling, by the way. It is used.

ransford:
I didn’t want to remove a real dependency so I grepped for “ephem” in the
entire source tree: no matches. Did I miss something?

DDevine:
I thought so at first, but you didn't.
However, it needs to be documented somewhere that for Solar schedules to work then ephem (and cpython) is needed.

thedrow:
Is ephem a dependency for Celery or django-celery-beat?

DDevine:
It's a Celery dependency, installed an extra (ie. pip install celery[solar] IIRC).

However, I don't know if things will crash if somebody creates a Solar based task in Django Admin when ephem is not installed.

PR #98, opened 2017-11-15
Adds support for message headers for periodic tasks

Viatrak:
This PR adds support for bundling message headers along with the task that will then get passed along during execution.

An example use case is with tenant-schemas-celery. TSC expects certain headers to be set in order to perform a task in a specific schema (tenant). By allowing headers to be saved with the periodic task, a recurring task can be created and used with tenant schemas celery.


Viatrak:
Merged to latest master. Added tests. Made everything green.

Viatrak:
The migrations were regenerated for this PR during the updates yesterday and it is compatible with master.

I tested with both a fresh DB migrate, and a DB migrate without this migration, and then migrate again with this migration made and added.

Is there some other way you'd like to see the migration or am I missing something here?

Thanks

auvipy:
there are 8 migrations, you could cross check your migrations with existing ones

Viatrak:
I am not quite clear what you're getting at. The naming and dependencies were generated by Django automatically. There are multiple of the same numbered non-conflicting migrations currently in the project. If it gets manually named as 0009_xyz.py there will be a conflict with multiple leaf nodes and will require some kind of migration merge migration. Is that a better route?

Viatrak:
OK, I think I have addressed your concerns with my latest push. I pushed migration changes that now have new numbering (0009_xyz) and merges the two migration dependency leafs together by adding an additional dependency in the migration. It looks when the periodic task priority pull request was merged in it did not add an appropriate migration.

auvipy:
@tinylambda could you have a look?

Viatrak:
Following up on this while it's still fresh -- are there any unaddressed issues?

Thanks

Viatrak:
Hi, any update on the status of this PR?

Thanks

PR #103, opened 2017-12-10
add on_delete=CASCADE on models PeriodicTask

mutalimov95:
to Django 2.0

TypeError: __init__() missing 1 required positional argument: 'on_delete'


vstoykov:
On delete is already added to these fields. See the error in travis and look again in the code.

PR #123, opened 2018-02-09
Size CrontabSchedule fields for max length values.

ShaheedHaque:
As per issue #122, it turns out that several of the fields backing a CrontabSchedule were too small to accomodate a maximum-length legal value. This PR addresses that.

Supporting tests are included. Please consider merging.


thedrow:
We need a migration file. Please run `python manage.py makemigrations` and add the result to this PR.

thedrow:
Thanks :)

PR #136, opened 2018-03-23
Improvement: Adds field PeriodicTask.origin_key to sync tasks defined in source code with settings.beat_schedule

pztrick:
This adds a field `PeriodicTask.origin_key` which indicates if a task database row was populated dynamically or from source code (e.g. `settings.CELERYBEAT_SCHEDULE` or `settings.beat_schedule` in `4.1+`).

If an entry is removed from `settings.CELERYBEAT_SCHEDULE` it will also delete any previously populated database row for `PeriodicTask`.

E.g.

```
# before
CELERYBEAT_SCHEDULE = {
    'test-sms': {
        'task': 'app.tasks.test_sms',
        'schedule': datetime.timedelta(seconds=60),
        'args': None
    },
}

# after
CELERYBEAT_SCHEDULE = {
    # 'test-sms': {
    #     'task': 'app.tasks.test_sms',
    #     'schedule': datetime.timedelta(seconds=60),
    #     'args': None
    # },
}
```

After commenting out the above, celery beat would purge the task from database and also emit this logger statement:

```
[2018-02-09 13:51:48,345: WARNING/Beat] Purged periodic tasks [origin_key=beat_schedule]: test-sms
```

---

I also updated the scheduler to only ingest one and only one schedule when reading `settings.CELERYBEAT_SCHEDULE`. The old behavior was that if you re-wrote an interval schedule to a cron schedule, it would remain associated with the old (now deleted) interval schedule and be called for both schedules. E.g.:

```
# old
CELERYBEAT_SCHEDULE = {
    'test-sms': {
        'task': 'app.tasks.test_sms',
        'schedule': datetime.timedelta(seconds=60),
        'args': None
    },
}

# new
CELERYBEAT_SCHEDULE = {
    'test-sms': {
        'task': 'app.tasks.test_sms',
        'schedule': crontab(minute='22', hour='14', day_of_week='*'),
        'args': None
    },
}
```

### DIFF
--- a/django_celery_beat/migrations/0007_periodictask_origin_key.py
+++ b/django_celery_beat/migrations/0007_periodictask_origin_key.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('django_celery_beat', '0006_auto_20180210_1226'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='periodictask',
+            name='origin_key',
+            field=models.CharField(
+                blank=True,
+                default=None,
+                max_length=200,
+                null=True,
+                verbose_name='origin key'
+            ),
+        ),
+    ]
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -142,12 +142,19 @@ class ModelEntry(ScheduleEntry):
                        args=None, kwargs=None, relative=None, options=None,
                        **entry):
         model_schedule, model_field = cls.to_model_schedule(schedule)
+        schedules = {
+            'interval': None,
+            'crontab': None,
+            'solar': None
+        }
+        schedules[model_field] = model_schedule
+
+        entry.update(**schedules)
         entry.update(
-            {model_field: model_schedule},
             args=dumps(args or []),
             kwargs=dumps(kwargs or {}),
-            **cls._unpack_options(**options or {})
         )
+        entry.update(**cls._unpack_options(**options or {}))
         return entry
 
     @classmethod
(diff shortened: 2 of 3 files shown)


django_celery_beat/schedulers.py
```diff
@@ -142,12 +142,19 @@ def _unpack_fields(cls, schedule,
...
         model_schedule, model_field = cls.to_model_schedule(schedule)
+        schedules = {
+            'interval': None,
+            'crontab': None,
+            'solar': None
+        }
+        schedules[model_field] = model_schedule
+
+        entry.update(**schedules)
         entry.update(
```
liquidpele: So this will run two queries instead of one now?  It would be nicer to keep it as one.

django_celery_beat/schedulers.py
```diff
@@ -248,18 +256,36 @@ def sync(self):
...
                 logger.error(ADD_ENTRY_ERROR, name, exc, entry_fields)
+        if origin_key:
+            # delete database-persisted periodic tasks that
+            # are no longer declared in source code
+            existing_task_instances = PeriodicTask.objects.filter(
+                origin_key=origin_key
+            )
+            existing_tasks = set(
+                map(lambda x: x.name, existing_task_instances)
+            )
```
liquidpele: Could replace a lot of logic here with just this I think: 

    tasks_to_purge = PeriodicTask.objects.filter(
        origin_key=origin_key).exclude(name__in=mapping.keys())
    if tasks_to_purge:
        tasks_to_purge.delete()
        log.warn(...)

django_celery_beat/migrations/0007_periodictask_origin_key.py
```diff
@@ -0,0 +1,25 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('django_celery_beat', '0006_auto_20180210_1226'),
```
liquidpele: Other migrations has been added, you'll need to rebase this to be the next one in the list based on what's in the master branch.

PR #141, opened 2018-04-24
run selected tasks in right queue

pualien:
changed `task.delay` to `task.apply_async` to run tasks in selected queues when specified in PeriodicTask queue charfield


liquidpele:
Thanks @pualien for your help!

PR #160, opened 2018-09-09
Warn user if task selected to run was not found

iAnanich:


iAnanich:
If user (admin) is trying to run periodic task that, for example, does not exist anymore, server will return 500, because it tries to get (and call) `delay` method of `None` object, so AttributeError gets raised. 
I added some checks so instead of server error user will get friendly message.

iAnanich:
about tests: one simple exception test would be enough, isn't it?

auvipy:
please proceed with that and try to make everything green!

auvipy:
also mention the issue number you are trying to fix

liquidpele:
@iAnanich hey there, will you be able to fix up your PR?  First you should merge latest master branch into your own branch as we have fixed some test issues, so that will tell us if your branch is actually passing the tests or not.  Plus, you'll have to fix the conflicts when you do so.  Thanks!  :)

iAnanich:
@liquidpele merged and fixed conflicts. Now it must work well. Even tests are passing )
But haven't tried it (after update) in action yet.

liquidpele:
@iAnanich looks like you have flake8 errors.

iAnanich:
@liquidpele yep, just fixed it

Issue #172, opened 2018-10-10
Can't create task if interval or solar schedule is selected


lucasdavila86:
I'm only able to a create scheduled task if I use a crontab schedule. I think this was introduced in this line
https://github.com/celery/django-celery-beat/compare/v1.1.1...master#diff-eeed49a50aa95de80188931bb26bee4eR328

Maybe this should be the solution
```
if self.interval and self.solar:
    raise ValidationError({
        'solar': [err_msg]
    })
```

zuoakang:
**When I chose Interval, why do I need to remind me to set Solar?**

> env
- python 3.6
- django_celery_beat==1.2.0

![image](https://user-images.githubusercontent.com/22383362/47482000-ea5d1500-d867-11e8-8152-d4c2cf213d66.png)

legshort:
I also did downgraded to 1.1.1 too

auvipy:
try 1.4 release

PR #173, opened 2018-10-10
raised ValidationError when interval or solar, should be interval and solar

lucasdavila86:
updated the validate_unique method on the PeriodicTask model
raise error when interval AND solar
issue #172


tinylambda:
> updated the validate_unique method on the PeriodicTask model
> raise error when interval AND solar
> issue #172

I rewrote the method `validate_unique `

https://github.com/celery/django-celery-beat/pull/178

PR #178, opened 2018-10-12
Implement validate_unique in a more elegant way

tinylambda:
The original implementation is complex and error-prone, so I refactor the code and solve a validation bug in the original implementation.

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -321,7 +321,7 @@ class PeriodicTask(models.Model):
                                    for schedule_type in schedule_info
                                    if schedule_info[schedule_type]]
 
-        if not any(selected_schedule_types):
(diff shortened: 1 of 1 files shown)


django_celery_beat/schedulers.py
```diff
@@ -252,19 +252,20 @@ def reserve(self, entry):
...
-            with transaction.atomic():
-                while self._dirty:
-                    try:
-                        name = self._dirty.pop()
-                        _tried.add(name)
-                        self.schedule[name].save()
-                    except (KeyError, ObjectDoesNotExist):
-                        pass
+
+            while self._dirty:
```
auvipy: isn't this part redundant?

tinylambda: Sorry, I will create a new PR right now

PR #179, opened 2018-10-12
Implement validate_unique in a more elegant way and solve a validatio…

tinylambda:
The original implementation is a little complex and error-prone, so I refactor the code and solve a validation bug in the original implementation.

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -317,9 +317,8 @@ class PeriodicTask(models.Model):
         schedule_info = {'interval': self.interval,
                          'crontab': self.crontab,
                          'solar': self.solar}
-        selected_schedule_types = [schedule_type
-                                   for schedule_type in schedule_info
-                                   if schedule_info[schedule_type]]
+        schedule_types = ['interval', 'crontab', 'solar']
+        selected_schedule_types = [s for s in schedule_types if getattr(self, s)]
 
         if len(selected_schedule_types) == 0:
             raise ValidationError({
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -316,7 +316,8 @@ class PeriodicTask(models.Model):
         super(PeriodicTask, self).validate_unique(*args, **kwargs)
 
         schedule_types = ['interval', 'crontab', 'solar']
-        selected_schedule_types = [s for s in schedule_types if getattr(self, s)]
+        selected_schedule_types = [s for s in schedule_types
+                                   if getattr(self, s)]
 
         if len(selected_schedule_types) == 0:
             raise ValidationError({
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -314,21 +314,26 @@ class PeriodicTask(models.Model):
 
     def validate_unique(self, *args, **kwargs):
         super(PeriodicTask, self).validate_unique(*args, **kwargs)
-        if not self.interval and not self.crontab and not self.solar:
+        schedule_info = {'interval': self.interval,
+                         'crontab': self.crontab,
+                         'solar': self.solar}
+        selected_schedule_types = [schedule_type
+                                   for schedule_type in schedule_info
+                                   if schedule_info[schedule_type]]
+
+        if len(selected_schedule_types) == 0:
             raise ValidationError({
                 'interval': [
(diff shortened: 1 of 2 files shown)


auvipy:
check the flake8 error

django_celery_beat/models.py
```diff
@@ -314,21 +314,26 @@ class Meta:
...
         super(PeriodicTask, self).validate_unique(*args, **kwargs)
-        if not self.interval and not self.crontab and not self.solar:
+        schedule_info = {'interval': self.interval,
+                         'crontab': self.crontab,
+                         'solar': self.solar}
+        selected_schedule_types = [schedule_type
+                                   for schedule_type in schedule_info
+                                   if schedule_info[schedule_type]]
+
+        if len(selected_schedule_types) == 0:
```
auvipy: could you explain the logic of if len(selected_....) == 0: ?

django_celery_beat/models.py
```diff
@@ -314,21 +314,26 @@ class Meta:
...
         err_msg = 'Only one of interval, crontab, or solar must be set'
-        if (self.interval and self.crontab) or (self.crontab and self.solar):
-            raise ValidationError({
-                'crontab': [err_msg]
-            })
-        if self.interval or self.solar:
-            raise ValidationError({
-                'solar': [err_msg]
-            })
+        if len(selected_schedule_types) > 1:
```
auvipy: also this logic?  ....>1:

django_celery_beat/models.py
```diff
@@ -314,21 +314,26 @@ class Meta:
 
     def validate_unique(self, *args, **kwargs):
         super(PeriodicTask, self).validate_unique(*args, **kwargs)
-        if not self.interval and not self.crontab and not self.solar:
+        schedule_info = {'interval': self.interval,
+                         'crontab': self.crontab,
+                         'solar': self.solar}
+        selected_schedule_types = [schedule_type
+                                   for schedule_type in schedule_info
+                                   if schedule_info[schedule_type]]
```
rsichny: maybe something more obvious like `selected_schedule_types = [s for s in ['interval', 'crontab', 'solar'] if getattr(self, s)]`?

django_celery_beat/models.py
tinylambda: `len(selected_schedule_types) == 0` means that no schedule type is selected , and there should be one and only one schedule type selected.

django_celery_beat/models.py
tinylambda: There should be one and only one schedule type selected. `len(selected_schedule_types)` means that more than one schedule type is selected, so raise ValidationError

django_celery_beat/models.py
tinylambda: > maybe something more obvious like `selected_schedule_types = [s for s in ['interval', 'crontab', 'solar'] if getattr(self, s)]`?

yes, more clear and beautiful. update it !

t/unit/test_admin.py
```diff
@@ -51,3 +55,20 @@ def test_toggle_action_all_disabled(self):
...
+            periodic_task.validate_unique()
+
+        periodic_task.crontab = crontab_schedule
+        ret_value = periodic_task.validate_unique()
+        self.assertIsNone(ret_value,
+                          'validate_unique should return None')
+
+        periodic_task.interval = interval_schedule
+        with self.assertRaises(ValidationError):
+            periodic_task.validate_unique()
```
rsichny: i think it would be useful to add separate test cases for each type:
```
from itertools import permutations

def test_validate_unique_raises_if_schedule_not_set(self):
    with self.assertRaises(ValidationError):
        PeriodicTask().validate_unique()

def test_validate_unique_raises_if_for_multiple_schedules(self):
    schedules = [
        ('crontab', CrontabSchedule()),
        ('interval', IntervalSchedule()),
        ('solar', SolarSchedule()),
    ]
    for options in permutations(schedules, 2):
        with self.assertRaises(ValidationError):
            PeriodicTask(**options).validate_unique()

def test_validate_unique_not_raises(self):
    PeriodicTask(crontab=CrontabSchedule()).validate_unique()
    PeriodicTask(interval=IntervalSchedule()).validate_unique()
    PeriodicTask(solar=SolarSchedule()).validate_unique()
```

PR #183, opened 2018-10-13
fix master migrations

### LINKED ISSUE #177: Migration rename leads to a dependency error when upgrading
Migration with sequence number 0006 [0006_auto_20180210_1226](https://github.com/celery/django-celery-beat/blob/v1.1.1/django_celery_beat/migrations/0006_auto_20180210_1226.py) in v1.1.1 was renamed to [0006_auto_20180322_0932](https://github.com/celery/django-celery-beat/blob/v1.2.0/django_celery_beat/migrations/0006_auto_20180322_0932.py) in v1.2.0 (and the contents changed).

This causes the following error with our own migrations pointing to the v1.1.1 migration.
```
django.db.migrations.exceptions.NodeNotFoundError: Migration myapp.0001_initial dependencies reference nonexistent parent node ('django_celery_beat', '0006_auto_20180210_1226')
```

What was the reasoning behind renaming the migration instead of creating a new one? #99 introduced the renamed migration.
Since this change was introduced in a semver wise non breaking version bump, should we create a fix for this, restoring the old 0006 migration and adding the new migrations since v1.1.1 from 0007 and onward?

rsichny:
fixes #177 

*this may ruin migrations tree for people who constantly use master branch*

tested scenarios:
- 1.1.0 -> latest
- 1.1.0 -> 1.1.1 -> latest
- 1.1.0 -> 1.2.0 -> latest
- 1.1.1 -> 1.2.0 -> latest
- current master -> latest
- makemigrations on latest code does not detect any changes


robvdl:
You can't migrate back anymore after applying the current migrations, it completely breaks the migration state if you do. None of the tested scenarios listed there try migrating backwards.

PR #185, opened 2018-10-17
After changing the PeriodicTask for one-off task we should mark it ch…

tinylambda:
If a PeriodicTask's enabled is set to False, it should be removed from the current schedule or else the is_due method will always return at the first conditions test which is in a bad state.

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -97,14 +97,14 @@ class ModelEntry(ScheduleEntry):
 
     def is_due(self):
         if not self.model.enabled:
-            return False, 5.0   # 5 second delay for re-enable.
+            return schedules.schedstate(False, 5.0)   # 5 second delay for re-enable.
 
         # START DATE: only run after the `start_time`, if one exists.
         if self.model.start_time is not None:
             if maybe_make_aware(self._default_now()) < self.model.start_time:
                 # The datetime is before the start date - don't run.
                 _, delay = self.schedule.is_due(self.last_run_at)
-                return False, delay  # use original delay for re-check
+                return schedules.schedstate(False, delay)  # use original delay for re-check
 
         # ONE OFF TASK: Disable one off tasks after they've ran once
         if self.model.one_off and self.model.enabled \
@@ -113,7 +113,7 @@ class ModelEntry(ScheduleEntry):
             self.model.total_run_count = 0  # Reset
             self.model.no_changes = False  # Mark the model entry as changed
             self.model.save()
-            return False, None  # Don't recheck
+            return schedules.schedstate(False, None)  # Don't recheck
 
         return self.schedule.is_due(self.last_run_at)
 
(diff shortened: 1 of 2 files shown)


tinylambda:
If we had this type of tasks in schedule, it will be always the first of the _heap, so other tasks will not get a chance to be executed

auvipy:
can you demonstrate a failing test case?

tinylambda:
> can you demonstrate a failing test case?

1. Add three tasks at the admin site
2. Set one of them to be one-off task
3. Start Celery Beat 

And wait for the first run of the one-off task, after that other two tasks will never be executed

tinylambda:
Disabled task should be deleted from the schedule

tinylambda:
In the implementation of `tick` method in  `celery.beat module`, if is_due is `False`, the `_heap` in the Scheduler will not be touched, and will retrieve exactly the same entry from the `_heap` in the following iterations, which is a infinite loop, it's a bad state.

So we should avoid keeping disabled tasks in the schedule, if a task is set to disabled, it means that it should be deleted from the schedule too.

tinylambda:
def test_heap_always_return_the_first_item2(self):
        interval = 10

        s1 = schedule(timedelta(seconds=interval))
        m1 = self.create_model_interval(s1, enabled=False)  # return False, 5.0 in is_due
        m1.last_run_at = self.app.now() - timedelta(seconds=interval+2)
        m1.save()
        m1.refresh_from_db()

        s2 = schedule(timedelta(seconds=interval))
        m2 = self.create_model_interval(s2, enabled=True)
        m2.last_run_at = self.app.now() - timedelta(seconds=interval+1)
        m2.save()
        m2.refresh_from_db()

        e1 = EntryTrackSave(m1, self.app)
        e2 = EntryTrackSave(m2, self.app)  # because the disabled task e1 is run first, this will never be executed

        self.s.schedule.clear()
        self.s.schedule[e1.name] = e1
        self.s.schedule[e2.name] = e2
        self.s._heap = None

        tried = set()
        for _ in range(len(self.s.schedule) * 16):
            tick_interval = self.s.tick()
            if tick_interval and tick_interval > 0.0:
                tried.add(self.s._heap[0].entry.name)
                time.sleep(tick_interval)
                if self.s.should_sync():
                    self.s.sync()

        assert len(tried) == 1 and tried == set([e1.name])

auvipy:
why not add the test example you demonstrate to the pr?

tinylambda:
In fact, I don't think we need a test case here, because it's obviously a logic bug and we can fix it by just adding just one line of code.

though I am a fan of test-driven development, we should not go too far.

bodgerbarnett:
I'm not sure the test case above actually replicates this problem. It tests that the first task (e1) is tried, whereas surely you want it to test that the second task (e2) is also tried, no? Or have I misunderstood this issue?

tinylambda:
> I'm not sure the test case above actually replicates this problem. It tests that the first task (e1) is tried, whereas surely you want it to test that the second task (e2) is also tried, no? Or have I misunderstood this issue?

Yes, you are right, it's just a simulation of the problem, I don't think we need a test case like this, or another way of writing the test

PR #233, opened 2019-03-15
make last_run_at tz aware before passing to celery

### LINKED ISSUE #211: Issue with Django USE_TZ=False and TIME_ZONE set
The situation:
celery 4.2.1
django-celery-beat 1.4.0
Django 2.1.3
Python 3.7
Postgres 9

django settings.py:
TIME_ZONE = 'Europe/Moscow'
USE_TZ = False

> celery report
software -> celery:4.2.1 (windowlicker) kombu:4.2.2 py:3.7.1
billiard:3.5.0.5 redis:3.0.1
platform -> system:Linux arch:64bit imp:CPython
loader -> celery.loaders.app.AppLoader
settings -> transport:redis results:django-cache
result_backend: 'django-cache'
task_serializer: 'pickle'
accept_content: ['json', 'pickle']
result_serializer: 'pickle'
timezone: 'Europe/Moscow'
result_expires: datetime.timedelta(days=14)
task_track_started: True
broker_url: 'redis://redis:6379/1'
worker_log_color: False
worker_max_tasks_per_child: 50
worker_prefetch_multiplier: 1

The project is using naive datetime and i know what i'm doing.

The problem is that task with period of 1 minute schedule is started constantly by beat in loop.

> beat_1 | [2018-12-19 09:00:54,575: DEBUG/MainProcess] .tasks.server_tasks.tasks_scheduler sent. id->03755bc5-3cfc-408d-928c-69092e8f09cb
beat_1 | [2018-12-19 09:00:54,586: INFO/MainProcess] Scheduler: Sending due task (tasks.server_tasks.tasks_scheduler)
beat_1 | [2018-12-19 09:00:54,588: DEBUG/MainProcess] .tasks.server_tasks.tasks_scheduler sent. id->3ecb6d84-ec1a-4684-bece-90d10c4f2bd8
beat_1 | [2018-12-19 09:00:54,600: INFO/MainProcess] Scheduler: Sending due task (.tasks.server_tasks.tasks_scheduler)
beat_1 | [2018-12-19 09:00:54,602: DEBUG/MainProcess] .tasks.server_tasks.tasks_scheduler sent. id->5981449a-2626-48f1-90ab-0ff6d17ba5de
beat_1 | [2018-12-19 09:00:54,611: INFO/MainProcess] Scheduler: Sending due task (.tasks.server_tasks.tasks_scheduler)

In postgres times stored as UTC and my connection has right timezone set.
My datetime fields is always has local value in local timezone as set in settings.py.

After debugging i see that problem is that celery.beat.Scheduler.is_due always returns negative values and task is started at every tick.

I see that few function dealing with time conversion that beat is using undercover. It is:
celery.utils.time.make_aware
celery.utils.time.maybe_make_aware
django_celery_beat.utils.make_aware

django_celery_beat.utils.make_aware - works fine, depends on USE_TZ. If USE_TZ is False it doing nothing, if True it makes assumption that naive time is UTC and converts it to local timezone.

celery.utils.time.maybe_make_aware is used in celery.schedules.schedule.is_due with last_run_at.
last_run_at in my case is naive time already in local timezone! But celery.utils.time.maybe_make_aware thinks it is UTC time.

Django with USE_TZ=False imply that naive time is always local time.
But celery function celery.utils.time.maybe_make_aware thinks that naive time is always UTC time and makes wrong conversion.

I think core problem is that celery makes "wrong" assumption on naive time timezone, but never mind...

The fix, i think, is that django-celery-beat must take this into account and convert all naive time to UTC when passing to celery functions.

costela:
celery always makes times TZ-aware and will wrongly assume they are in UTC, even when the
dates from the django model have an (implicit) non-UTC TZ, like in cases where `USE_TZ=False` and `TIME_ZONE` is set to something other than `UTC`.

this should fix #211

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -93,12 +93,7 @@ class ModelEntry(ScheduleEntry):
         if not model.last_run_at:
             model.last_run_at = self._default_now()
 
-        last_run_at = model.last_run_at
-
-        if getattr(settings, 'DJANGO_CELERY_BEAT_TZ_AWARE', True):
-            last_run_at = make_aware(last_run_at)
-
-        self.last_run_at = last_run_at
+        self.last_run_at = model.last_run_at
 
     def _disable(self, model):
         model.no_changes = True
@@ -133,7 +128,7 @@ class ModelEntry(ScheduleEntry):
             self.model.save()
             return schedules.schedstate(False, None)  # Don't recheck
 
-        return self.schedule.is_due(self.last_run_at)
+        return self.schedule.is_due(make_aware(self.last_run_at))
 
     def _default_now(self):
         # The PyTZ datetime must be localised for the Django-Celery-Beat
--- a/django_celery_beat/utils.py
+++ b/django_celery_beat/utils.py
@@ -21,6 +21,10 @@ def make_aware(value):
         # then convert to the Django configured timezone.
         default_tz = timezone.get_default_timezone()
         value = timezone.localtime(value, default_tz)
+    else:
+        # naive datetimes are assumed to be in local timezone.
+        if timezone.is_naive(value):
+            value = timezone.make_aware(value, timezone.get_default_timezone())
     return value
 
 
--- a/t/unit/test_schedulers.py
+++ b/t/unit/test_schedulers.py
@@ -9,7 +9,8 @@ from itertools import count
 
 from django.contrib.admin.sites import AdminSite
 from django.contrib.messages.storage.fallback import FallbackStorage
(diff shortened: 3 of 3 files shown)


costela:
Build is actually green, with the exception of `linkcheck` and `apicheck`, which got connection errors to `docs.celeryproject.org`. Could someone please kick the build jobs?

auvipy:
restarted the build

costela:
@auvipy thanks!
There's still something weird going on with linkcheck and apicheck. :thinking:

auvipy:
You could check if time permits

PR #239, opened 2019-04-07
brought new django and python versions to maatix

auvipy:


liquidpele:
Yea, it looks like by default Travis uses ubuntu 14.04, which is weird since that's so damn old.  We should probably change to use xenial (16.04) for everything.  Online discussions/bitching about them not yet supporting bionic (18.04) tell me it'll be a while for that still unfortunately, and 14.04 goes EOL this month, so I guess we need to just use xenial.

This stackoverflow answer has an example of updating to xenial for all tests: 
https://stackoverflow.com/questions/55521832/update-sqlite-in-travis-ci

auvipy:
Xenial default and trusty for specific versions if needed. gonna update soon

liquidpele:
We should probably also test pypy3.5 since just pypy is based on 2.7

https://docs.travis-ci.com/user/languages/python/#pypy-support

liquidpele:
I think I ended up doing this stuff as part of my PR in order to fix that weird sqlite bug, should this be closed or was there other stuff you wanted to do too?

liquidpele:
Closing this for now, just reopen if there was more you wanted to do :)

PR #243, opened 2019-04-26
Fix up tests

liquidpele:
1. add test runs for python 3.7, remove 3.4
2. fix 2.7 tests to not use celery versions that don't support it
3. add pytest-timeout to avoid hanging certain tests when broker is not running


requirements/python2.txt
```diff
@@ -0,0 +1,3 @@
+# celery stops supporting python2 at 4.3.0
+# thus for python2 tests we can't use latest master anymore
+celery<4.3.0
```
auvipy: this will be less than 5.0, celery didn't stop python 2 at 4.3. the master will support py2 again and python 5 will be moved to celery 5 branch

requirements/test.txt
```diff
@@ -1,4 +1,5 @@
 case>=1.3.1
-pytest>=3.0
 pytest-django
 pytz>dev
+pytest<4.0.0
```
auvipy: we should use latest pytest as celery also use that.

requirements/python2.txt
liquidpele: The dependency tree for 4.3.0 installs the alpha of 5.0 of vine, which then causes the tests to bomb with:

```
    class Thenable(Callable, metaclass=abc.ABCMeta):  # pragma: no cover
                                     ^
SyntaxError: invalid syntax
```

liquidpele: oooh, wait, I think I see what you're saying.  Okay, one sec.

auvipy: you can pin vine

requirements/test.txt
liquidpele: I was getting errors from this with 4.0+

```
AttributeError: 'TestCaseFunction' object has no attribute 'get_marker'
```

auvipy: i will push pytest update later. keep this less then 4 for now.

requirements/python2.txt
liquidpele: I've updated it to use celery<5.0.0

liquidpele:
@auvipy Updated with working python 3.7 and all tests passing.   

Notes:
1. I temporarily disabled the doc jobs due to an upstream bug, which I opened here:  https://github.com/celery/sphinx_celery/issues/9
2. flake8 was newer on python 3.5 and caught an extra item, so I fixed that. 
3. python3.7 isn't fully supported by travis yet, so it required installing xenial and then updating erlang to get rabbitmq to not bomb out for the tests.

Issue #210, opened 2018-12-19
upgraded from beat 1.0.1 to 1.4.0 and beat has died twice in two days


robvdl:
We recently upgraded a UAT system from Celery 4.1.1 to 4.2.1 and Beat to 1.4.0

At first glance everything was fine, but twice now has beat process randomly just died and we've had to restart it.

The reason for the upgrade was to switch to the redis-py library version 3, with celery 4.1 we had to pin redis-py to version 2 since the recent redis-py 3 release broke things initially.

Anyway, beat dying has never happened before until we upgraded it, I'll see if I can get some additional information but in the mean time we have to do a rollback to Celery 4.1.1 and Beat 1.0.1

robvdl:
The issue with the broken migration 0005 shows up if you try to go from 0006 to 0005:

./manage.py migrate django_celery_beat 0005

CommandError: More than one migration matches '0005' in app 'django_celery_beat'. Please be more specific.

That is because 0005 is that squash migration, it seems to have messed with the order.

robvdl:
There are multiple 0006 and multiple 0005 migrations, this thing is a real mess and should be resolved.

It looks like someone did try to resolve it, but you cannot reverse migrate or jump to a specific migration because so many re-used numbers.

robvdl:
I also tried to empty all the beat related tables in the Django admin first before migrating back to 0001 and even then it blows up unfortunately.

psycopg2.ProgrammingError: column django_celery_beat_periodictask.solar_id does not exist
LINE 1: ..., "django_celery_beat_periodictask"."crontab_id", "django_ce...

Basically I'm going to have to go into the db and delete the tables and reset the migration state of the django_celery_beat app and just tart again, it looks like the migrations are broken and it simply is not possible to migrate back.

justmobilize:
I'm happy to help fix the migrations. Version `1.3.0` and `1.4.0` should be pulled because once they are installed it will be difficult to get fixed. Squashing migrations doesn't work for a package because you never know what someone is on. It's different if you are in control of all the places the code is running.

robvdl:
Thanks, in the mean time I've manually rolled back to beat version 1.0.1 and rolled back celery to 4.1.1, because this is a live production system it can't really go down.

I'm going to do some experimentation in the new year to see if I can just upgrade celery but hold back beat for now, it would be nice to find out why beat keeps dying on me but it's a production system for a client so I can't really afford to mess with it too much.

auvipy:
[...]

PR #244, opened 2019-04-29
Migration tests, fix migration bug, and clean up test matrix.  

liquidpele:
In this PR:

1. Set up tests to upgrade from 1.1.1, 1.20, 1.3.0, 1.4.0.   After upgrade, run migrations back to 0001 and then forward again to ensure no issues.  
2. Add a migration to remove bad migration history caused by the migration squash a while back.  This was only causing an error when ugprading from 1.3.0 and then going backward again, caught by the new tests mentioned above. 
3. Update all test VMs to xenial (ubuntu 6.04).  Add erlang fixing code to a central location for all builds. 
4. Ensure we get latest sqlite due to a very strange timing bug with sqlite and migrations that would cause migrations to randomly fail. 
5. Add python 3.6, and django 2.1 to text matrix.


justmobilize:
@liquidpele are you going to do this for all versions?

liquidpele:
Yes, just playing with getting one working first... tox and travis-ci are finicky AF.

justmobilize:
I recommend looking here: #210 and not squashing the migration. You will have users in different spots and this will never work for all

auvipy:
Seems you are going to be martyred for fixing this :dagger:

liquidpele:
Ha.  Such an annoying issue.  I can't reproduce it anywhere except the damn travis-ci environment!

liquidpele:
@auvipy @justmobilize 

I think this is finally done!  Apparently the issue I was debugging was a but with django migrations and older sqlite.... basically it would just randomly fail saying that a field didn't exist, but re-running the same test again would make it pass, and I couldn't reproduce it locally.  Updating all tests to run on xenial seems to have stopped the failures.

Issue #217, opened 2019-01-10
Messed up migration files


justmobilize:
Hello! Are there plans to create a new version with corrected migrations? I would love to use some of the new features, but until the migrations are stable, I can't.

I would be happy to help if I can get commitment to the release being rolled out quickly.

tuky:
similarly to here https://github.com/EliotBerriot/django-dynamic-preferences/issues/23 would it be possible to limit the `max_length` to 191? my current solution to that involves monkey patching `CharField.__init__` which is not nice.

tuky:
would you welcome a PR which makes the max_length configurable like so? https://github.com/python-social-auth/social-app-django/blob/master/social_django/migrations/0003_alter_email_max_length.py

justmobilize:
@tuky FYI, that's dangerous, since you can't change it after the migration runs. If this was my repo, I would never support that

tuky:
I object, you can change the setting and run a custom sql migration, which realizes the new value.

justmobilize:
True, but many users don't understand how to use that and it can cause a lot of issues.

Either way, goal #1 should be to get the current migrations in a good place

auvipy:
doesn't the 1.4 release fix the migration problems?

tuky:
i've had no problems applying all the migrations apart from `max_length` being too high for MySQL with InnoDB and utf8mb4 (all widely recommended configuration options. well except MySQL ;-).

justmobilize:
It depends on what version you had installed. Migration files after a release shouldn't be edited since they are already being used

auvipy:
@tinylambda

moseb:
Migrations were sane at 1.4.0 — except maybe two 0006 files both depending on 0005 — but things have gone "sideways" after that on `master` regarding migrations (…). Before I go into detail let me quote the Django docs how the squashing process [was designed to work](https://docs.djangoproject.com/en/2.2/topics/migrations/#migration-squashing):


# Django's view on squashing migrations 

> The recommended process is to squash, keeping the old files, commit and release, wait until all systems are upgraded with the new release (or if you’re a third-party project, just ensure your users upgrade releases in order without skipping any), and then remove the old files, commit and do a second release.

Note the `your users upgrade releases in order without skipping any` part — something that most projects might have a hard time ensuring for (or requesting from) user. So squashing might not even work well for django-celery-beat's case _at all_.

It goes on saying:
[...]

PR #246, opened 2019-05-02
Fix database migrations

moseb:
Meant to address issue #217


auvipy:
Looking forward to it

django_celery_beat/migrations/0005_add_solarschedule_events_choices_squashed_0009_merge_20181012_1416.py
```diff
@@ -1,95 +0,0 @@
-# Generated by Django 2.1.2 on 2018-10-12 14:18
```
liquidpele: So you deleted the squashed migration here, but that shipped in 1.4.0.  I'm surprised that it didn't complain about it being in the migrations table later on upgrade, but I guess if it works it works /shrug

django_celery_beat/migrations/0006_periodictask_priority.py
```diff
@@ -7,7 +7,11 @@
 
 class Migration(migrations.Migration):
     dependencies = [
-        ('django_celery_beat', '0005_add_solarschedule_events_choices_squashed_0009_merge_20181012_1416'), # noqa
+        ('django_celery_beat', '0005_add_solarschedule_events_choices'),
+        ('django_celery_beat', '0006_auto_20180210_1226'),
+        ('django_celery_beat', '0006_auto_20180322_0932'),
+        ('django_celery_beat', '0007_auto_20180521_0826'),
+        ('django_celery_beat', '0008_auto_20180914_1922'),
```
liquidpele: Please add a comment here about why a 0006 depends on numbers after it, since that's super unconventional and looks like it's wrong at first glance if you didn't know the history here.

django_celery_beat/migrations/0011_fix_migration_history.py
```diff
@@ -10,22 +10,6 @@ class Migration(migrations.Migration):
...
-            or name = '0006_auto_20180210_1226'
-            or name = '0006_auto_20180322_0932'
-            or name = '0007_auto_20180521_0826'
-            or name = '0008_auto_20180914_1922'
-            )
-            """, None)],
-            reverse_sql=migrations.RunSQL.noop
-        ),
+        # operations removed, migration kept to keep external
+        # reverse dependencies intact
```
liquidpele: This migration can just be deleted, it never shipped as a version and has only existed for a few days.

justmobilize: You could also update it to remove `0005_add_solarschedule_events_choices_squashed_0009_merge_20181012_1416` so some don't have a ghost migration. Or keep `0005_add_solarschedule_events_choices_squashed_0009_merge_20181012_1416` and have it be empty

moseb: I only deleted `0005_add_solarschedule_events_choices_squashed_0009_merge_20181012_1416` altogether because I have not found any way to keep it: Not empty, not unmodified, nothing.

moseb: Will delete `0011`, okay

django_celery_beat/migrations/0006_periodictask_priority.py
moseb: Comment added. If it's not good enough, please propose some wording. Okay?

liquidpele: I'm not sure what BEGIN ex and END ex mean.  

Maybe say something like 

    # depends on higher numbers due to a squashed migration
    # that was later removed due to migration issues it caused

django_celery_beat/migrations/0011_fix_migration_history.py
moseb: Resolving, please re-open as needed

liquidpele:
@moseb looks like flake8 is failing now due to the comment... but I'm not sure what BEGIN ex and END ex mean. 

Maybe say something like 

    # depends on higher numbers due to a squashed migration
    # that was later removed due to migration issues it caused

django_celery_beat/migrations/0006_periodictask_priority.py
moseb: Comment added ([as suggested](https://github.com/celery/django-celery-beat/pull/246#issuecomment-489171503))

moseb:
Thanks!

liquidpele:
Awesome, thanks @moseb for the help!

PR #262, opened 2019-06-02
requirements: Add direct dependencies to Django and celery (fixes #258)

### LINKED ISSUE #258: setup.py does not pull in Django or Celery as direct runtime dependencies
Hi!

I noticed that `setup.py` uses `requirements/*.txt` for dependencies, but none of the ones used during `pip install django-celery-beat` pull in Django and/or Celery, directly. (Django is pulled in indirectly, though). Is that a conscious decision? If not, can they be added?

Best, Sebastian

hartwork:
To address #258

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -9,7 +9,6 @@ RUN apt-get update && apt-get install --yes --no-install-recommends \
         wait-for-it
 
 RUN pip3 install --user \
-        celery \
         django-createsuperuserwithpassword \
         psycopg2-binary
 
--- a/requirements/runtime.txt
+++ b/requirements/runtime.txt
@@ -1,0 +1,2 @@
+celery
+Django>=1.11.17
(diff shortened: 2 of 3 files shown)


requirements/default.txt
```diff
@@ -1,2 +1,4 @@
+celery
+Django>=1.11.17
```
liquidpele: You can't add them to default, or the tests will all fail.  The tests must install a particular version depending on what variables the test is running at the time, and pip complains if you mention the same package twice.  

What you'll probably want to do is update the setup.py file to add just "celery" or "celery<5.0.0" to the install requirements for python3 and python2 respectively.

hartwork: Alright, second try :smile:

hartwork:
> This won’t install celery 5.x with python 2 will it? 5.x is what drops py2 support.

Not as long as celery itself gets things right, I believe.

hartwork:
i.e. if they use `python-requires`:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

hartwork:
Conflicts are fixed now

liquidpele:
Cool, looks good to me then.

PR #269, opened 2019-06-15
Improve upon (currently dangerous) .from_schedule code

hartwork:
For your consideration :smile:


django_celery_beat/models.py
```diff
@@ -87,13 +87,7 @@ def from_schedule(cls, schedule):
...
                 'latitude': schedule.lat,
                 'longitude': schedule.lon}
-        try:
-            return cls.objects.get(**spec)
-        except cls.DoesNotExist:
-            return cls(**spec)
-        except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+        return cls.objects.get_or_create(**spec)[0]
```
liquidpele: This can still raise MultipleObjectsReturned, why is not handling that better?

django_celery_beat/models.py
```diff
@@ -138,6 +132,7 @@ class Meta:
         verbose_name = _('interval')
         verbose_name_plural = _('intervals')
         ordering = ['period', 'every']
+        unique_together = ('every', 'period')
```
liquidpele: What if I want two different jobs to run on the same interval?  This would disallow that, that seems like an issue.

django_celery_beat/models.py
```diff
@@ -149,13 +144,11 @@ def schedule(self):
...
-        except cls.DoesNotExist:
-            return cls(every=every, period=period)
-        except MultipleObjectsReturned:
-            cls.objects.filter(every=every, period=period).delete()
-            return cls(every=every, period=period)
+        spec = {
+            'every': every,
+            'period': period,
+        }
+        return cls.objects.get_or_create(**spec)[0]
```
liquidpele: Same question about MultipleObjectsReturned

django_celery_beat/models.py
```diff
@@ -174,6 +167,7 @@ class ClockedSchedule(models.Model):
     clocked_time = models.DateTimeField(
         verbose_name=_('Clock Time'),
         help_text=_('Run the task at clocked time'),
+        unique=True,
```
liquidpele: Why limit it like this?  Having two jobs run at the same time is completely valid.

django_celery_beat/models.py
```diff
@@ -202,13 +196,7 @@ def schedule(self):
...
         spec = {'clocked_time': schedule.clocked_time,
                 'enabled': schedule.enabled}
-        try:
-            return cls.objects.get(**spec)
-        except cls.DoesNotExist:
-            return cls(**spec)
-        except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+        return cls.objects.get_or_create(**spec)[0]
```
liquidpele: Same question about MultipleObjectsReturned

django_celery_beat/models.py
```diff
@@ -280,6 +268,8 @@ class Meta:
         verbose_name_plural = _('crontabs')
         ordering = ['month_of_year', 'day_of_month',
                     'day_of_week', 'hour', 'minute', 'timezone']
+        unique_together = ('minute', 'hour', 'day_of_week', 'day_of_month',
+                           'month_of_year', 'timezone')
```
liquidpele: Same issue about limiting by the schedule.

django_celery_beat/models.py
```diff
@@ -317,13 +307,7 @@ def from_schedule(cls, schedule):
...
                 'timezone': schedule.tz
                 }
-        try:
-            return cls.objects.get(**spec)
-        except cls.DoesNotExist:
-            return cls(**spec)
-        except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+        return cls.objects.get_or_create(**spec)[0]
```
liquidpele: Same question about MultipleObjectsReturned

django_celery_beat/models.py
hartwork: There is a uniqueness constraint covering all fields that go into `spec` so the database should not be able to contain more than a single match at any time.

django_celery_beat/models.py
hartwork: I think there is a mixup between jobs and schedules here: You assign the same "every 4 days" schedule to two distinct jobs and they both run at the time. That works before and after this pull request.

The current `cls.objects.filter(**spec).delete()` is effectively a unique constraint, except that it might cascade destroy data and does not prevent creation in the first place.

I think what the code needs is one of:
- a) A uniqueness constraint for _all_ schedules (as done in the pull request)
- b) A uniqueness constraint for _none_ of the schedules and no dangerous  `cls.objects.filter(**spec).delete()` either

What do you think?

django_celery_beat/models.py
hartwork: Please see https://github.com/celery/django-celery-beat/pull/269#discussion_r294082204

django_celery_beat/models.py
liquidpele: I see your point, but even the previous logic seems terrible tbh.  

As an example, lets say I have 2 different schedules for 2 different tasks, one running every sunrise and one running every sunset (just using solarschedule as the example but this applies to any of them).  Now, lets say that my boss comes over and says all tasks running at sunset should be changed to sunrise for only the next week for whatever business reason.  Instead of modifying a single schedule, I now have to go through and modify potentially hundreds of different periodic tasks to change their schedule... and then I have to remember which ones I changed so I can change them back later.  

It seems like we should be allowing duplicate schedules.  I'm really curious why this limitation ever existed in the first place?

liquidpele: I'm digging into where all from_schedule is used to see if we can sort this out some, feel free to do the same.

hartwork: > Now, lets say that my boss comes over and says all tasks running at sunset should be changed to sunrise for only the next week for whatever business reason. Instead of modifying a single schedule, I now have to go through and modify potentially hundreds of different periodic tasks to change their schedule... and then I have to remember which ones I changed so I can change them back later.

I was thinking of the same scenario earlier. I think this feature also has downsides: Once you have multiple schedules with the same values, you'll have a hard time telling them apart when creating new jobs.

Personally, I see the value in a unique constraint but also the pain with not supporting to change a schedule into something that already exists.

To fully support the scenario that you have in mind, it would take another layer of indirection, I think.

hartwork: > I'm digging into where all from_schedule is used to see if we can sort this out some, feel free to do the same.

There is a single place outside the test suite:
https://github.com/celery/django-celery-beat/blob/db68a4997420e141b002b555d0d58e98f32a6a4f/django_celery_beat/schedulers.py#L171

liquidpele: Yea, I'm reading through the celery scheduler code those inherit from to work out the logic on what scenarios this call stack actually happens.

hartwork:
Any ideas how to proceed with https://github.com/celery/django-celery-beat/pull/269#discussion_r294082204 ?

django_celery_beat/models.py
liquidpele: So, based on what I've looked through, the whole reason for all this is that when you use the django settings to set up the schedules, it deletes/inserts those into the DB.  However, instead of having a flag for "inserted from django conf" or whatever, they compare it to what's already in the DB and treat the schedule items as if they are unique constraints.  Still kind of playing around with this, but if that's the case then the deletes may be necessary for now, although there are probably several scenarios where things can get borked.  Still playing around with this though.

hartwork:
Ping

django_celery_beat/models.py
auvipy: what about using uniqueConstraint instead?

hartwork: [`UniqueConstraint`](https://docs.djangoproject.com/en/3.0/ref/models/constraints/#uniqueconstraint) was added in Django 2.2. while django-celery-beat says to support older versions than that according to https://github.com/celery/django-celery-beat/blob/4fff14fb15574b5265529d56dc31cef403e339cb/setup.py#L51-L55

Also, `UniqueConstraint` seems to be more verbose. Which feature of `UniqueConstraint` were you looking for that `unique_together` does not offer?

auvipy: > I think there is a mixup between jobs and schedules here: You assign the same "every 4 days" schedule to two distinct jobs and they both run at the time. That works before and after this pull request.
> 
> The current `cls.objects.filter(**spec).delete()` is effectively a unique constraint, except that it might cascade destroy data and does not prevent creation in the first place.
> 
> I think what the code needs is one of:
> 
>     * a) A uniqueness constraint for _all_ schedules (as done in the pull request)
> 
>     * b) A uniqueness constraint for _none_ of the schedules and no dangerous  `cls.objects.filter(**spec).delete()` either
> 
> 
> What do you think?

just curious if we add UniqueContraint Check here would that be any useful?

hartwork: Hi @auvipy I would need get back into the whole topic since it's been so long ago by now. I can do that work but I would some "guarantee" that we'll actually get somewhere with this in the next few weeks to not invest in a dead end. (In case you're a friend of voice meetings for stuff like that, we can definitely do a Jitsi or Google Meet on the topic if you like, but no pressure, just one more option.)

auvipy: Ok, I will release a new version tomorrow and do some more analysis to get back to you to reach a consensus

izimobil:
FWIW, here is my take on this:
(see also: https://github.com/celery/django-celery-beat/pull/389 that fixes the very dangerous `from_schedule` implementation)

The idea of adding unique constraints is good, but there should be a warning in the changelog because it's not fully backwards compatible, existing code may try to create duplicate schedules (which currently works) and get a IntegrityError exception with the constraints added.

Furthermore, because existing databases can already contain duplicate schedules, the from_schedule methods should not be modified.

hartwork:
Closing as superseded by #389. 

CC #322 as related.

auvipy:
I might tackle this in the future if needed.

hartwork:
> I might tackle this in the future if needed.

@auvipy I see two unresolved topics that this pull request documents:
- introducing uniqueness constraints and
- potential to use `get_or_create` to simplify code and increase atomicity.

There is no way forward with the uniqueness constraint that *doesn't*…

- require user intervention on setups with existing duplicates and
- have downsides as well

…as explained by @liquidpele at https://github.com/celery/django-celery-beat/pull/269/files#r294091489 .

For `get_or_create`, a uniqueness constraint on the involved keys [would be "needed"](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#get-or-create) to not keep creating duplicate entries. So I see a soft deependency on the uniqueness constraint topic here.

With where we are today on `master` after #389 has been merged, this pull request titled "Improve upon (currently dangerous) .from_schedule code" is not the best way to keep those two potential todos in my opinion, plus I would like to make myself optional in this equation (which is in some conflict with me being the author of the pull request) and keep my own list of open pull request at https://github.com/pulls organized, so I'll close my pull request here again and ask you to create a new dedicated issue to address the topic of adding uniqueness constraints if that's still of interest.  Thanks for your understanding :pray: 

Best, Sebastian

PR #273, opened 2019-06-20
clockedschedule.py: Re-use schedstate of Celery >=3.1.0 (fixes #267)

### LINKED ISSUE #267: [1.5.0] clockedschedule.py redefines celery.schedules.schedstate of Celery >=3.1.0?
Hi!

I noticed this code:
https://github.com/celery/django-celery-beat/blob/db68a4997420e141b002b555d0d58e98f32a6a4f/django_celery_beat/clockedschedule.py#L10

I would expect `from celery.schedules import schedstate` instead, which requires Celery >=3.1.0.

Is there need to support Celery <3.1.0? otherwise, I would suggest to turn this into an import, e.g. to be safe towards future use of `isinstance` that would fail for non-obvious reasons.

What do you think?

CC @yywing

hartwork:
Fixes #267

### DIFF
--- a/django_celery_beat/clockedschedule.py
+++ b/django_celery_beat/clockedschedule.py
@@ -6,8 +6,10 @@ from celery import schedules
 from celery.utils.time import maybe_make_aware
 from collections import namedtuple
 
-
-schedstate = namedtuple('schedstate', ('is_due', 'next'))
+try:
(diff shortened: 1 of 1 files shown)


liquidpele:
Looks fine.

Issue #274, opened 2019-07-01
a new PT with future start_time+long time interval schedule blocks the heap


paulhuo:
Hi, when I have some short time interval schedule tasks, and running normal, then created a 1-day interval schedule with future start_time=now+1minute. 
As expected, this periodtask has been popped out the heap, but didnt be sent to worker, the worse is:the heap seems blocked, the former short time interval schedule tasks no longer to be sent to worker.

seems the problem is:
last_run_at of a new task is set in ModelEntry
`        if not model.last_run_at:
            model.last_run_at = self._default_now()`
when schedule.is_due(last_run_at) is called,  it returns false because remaining(last_run_at with 1-day interval)>0, so the heap(in beat tick function) will not pop until this task comes 1-day latter, this blocks the other normal tasks.

btw: I traced the heap in beat tick function, the start_time works fine, so this problem is not like others with start_time.

* Celery Version: 4.3.0
* Celery-Beat Version: 1.5.0

scenario looks like this:
1. create a interval task A with 30 seconds, and start_time=now
2. create a interval task B with 1 day, and start_time=now+1 minute
3. A is sent to worker within 1 minute, after that it doesnt work

ggwakaka:
I have the same problem.
I find the reason that  1-day interval schedule next time is start time, because now < start_time when celery beat's populate_heap() call  ModelEntry's is_due().After a minute schedule was executed,  
tick will call the ModelEntry's is_due(), and return a day's timedelta because now > start_time.but 
time in event_t of 1-day interval also is the least value,so it is on the top of beat's heap.

I try to set last_run_at to start_time when start_time is not None and start_time > now when init ModelEntry:
`
        if not model.last_run_at:
            if model.start_time is not None and model.start_time > self._default_now():
                model.last_run_at = model.start_time
            else:
                model.last_run_at = self._default_now()
`
Then remove the code:
`
        if self.model.start_time is not None:
            now = self._default_now()
            if getattr(settings, 'DJANGO_CELERY_BEAT_TZ_AWARE', True):
                now = maybe_make_aware(self._default_now())

            if now < self.model.start_time:
                # The datetime is before the start date - don't run.
                # send a delay to retry on start_time
                delay = math.ceil(
                    (self.model.start_time - now).total_seconds()
                )
                return schedules.schedstate(False, delay)
`
It works for me.
I think you can have a trial.

paulhuo:
I think the key is how to set the default value of `last_run_at` when the task never run before.
[...]

PR #276, opened 2019-07-02
fix bug for #274

ggwakaka:
fix bug for #274 https://github.com/celery/django-celery-beat/issues/274#issue-462630011

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -93,6 +93,9 @@ class ModelEntry(ScheduleEntry):
         self.total_run_count = model.total_run_count
         self.model = model
 
+        if model.start_time is not None and model.start_time > self._default_now():
+            model.last_run_at = model.start_time
+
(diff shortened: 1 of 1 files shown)


django_celery_beat/schedulers.py
```diff
@@ -109,18 +112,18 @@ def is_due(self):
...
-            now = self._default_now()
-            if getattr(settings, 'DJANGO_CELERY_BEAT_TZ_AWARE', True):
-                now = maybe_make_aware(self._default_now())
+        #if self.model.start_time is not None:
+        #    now = self._default_now()
+        #    if getattr(settings, 'DJANGO_CELERY_BEAT_TZ_AWARE', True):
+        #        now = maybe_make_aware(self._default_now())
 
-            if now < self.model.start_time:
+        #    if now < self.model.start_time:
```
liquidpele: Seems more like the fix should be changing the < to <= here.

PR #288, opened 2019-11-15
Add rabbit mq container in dev einvironment

naggappan:
Adding Rabbit MQ container in dev environment , so that we can test the scheduler component is actually pushing the message or not. Or else beat component fails with error cannot connect to rabbitmq since it looks by default in localhost

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ services:
     depends_on:
       - base
       - postgres
+      - rabbit
     build:
       context: .
       dockerfile: docker/django/Dockerfile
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -29,14 +30,20 @@ services:
...
     entrypoint: ["/app/docker/celery-beat/entrypoint.sh"]
+    environment:
+        CELERY_BROKER_URL: 'amqp://guest:guest@rabbit:5672'
     command: ["python3", '-m', "celery", "-A", "mysite", "beat", "-l", "info", "--scheduler", "django_celery_beat.schedulers:DatabaseScheduler"]
     tty: true
     volumes:
       - './django_celery_beat/:/app/django_celery_beat/'
 
+  rabbit:
+      image: rabbitmq:3.8.0
```
auvipy: is 3.8.1 available?

naggappan: Yes it is available, I just took this version from celery project. Do we need to restrict to some version? can we put latest tag ? so that always the latest stable version will be available in dev environment.

naggappan:
UT cases are failing , I think already the master branch has some issue. Is someone working on this? I don't see any issue raised with respect to UT failures. Is someone looking into this issue? 

Also I see that previous PR got merged without the Travis CI pass, is it possible to merge without the UT pass?

docker-compose.yml
auvipy: ye can put latest tag

naggappan: I have added with latest tag.
@auvipy test case are failing, how to merge the code to master? I see that even it is failing for master branch or other PR as well

PR #294, opened 2019-12-03
Fix case where last_run_at=None and CELERY_TIMEZONE!=TIME_ZONE

marky1991:
In this PR, I'd like to fix a case where beat causes celery to misinterpret the hours of a crontab. The exact data scenario is given below. Skimming through the open issues, it's possible that there are other lurking scenarios that could have similar issues, but I'm pretty confident that this PR at least fixes this scenario. The ultimate conclusion is that celery's crontab.is_due must always be given a datetime in terms of CELERY_TIMEZONE, not TIME_ZONE and this change ensures that. 

My analysis of my particular scenario:

With CELERY_TZ_AWARE= False, USE_TZ=False, Django TIMEZONE="UTC", and CELERY_TIMEZONE="Los_Angeles"
If all last_run_at=None:
    django_celery_beat.ModelEntry is_due gets called by beat, and assumes last_run_at=_get_default_now().
        Because the TZ environment variable has been set to django's timezone, UTC, it returns a
        naive object that corresponds to the current time in utc.

    django_celery_beat.ModelEntry.is_due then calls make_aware on this, which just adds the UTC tzinfo to the datetime. (it uses the tz returned by
        django.utils.timezone.get_default_timezone(), which just returns the value of TIMEZONE)
    celery.schedules.crontab.is_due is next called and given this datetime, delegating to remaining_estimate.
    remaining_estimate then calls remaining_delta
    remaining_delta first gets now(), using CELERY_TIMEZONE, returning a tz-aware datetime, with tzinfo=los_angeles.
        It also calls maybe_make_aware on the last_run_ts object, but this function doesn't do anything if the datetime isn't naive
    Finally, remaining_delta compares the hour attribute of last_run_ts (which is in UTC) with the hours attribute of the crontab object, which is in terms of CELERY_TIMEZONE, not settings.TIME_ZONE. Because of this, the hour never matches and the job never fires.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -258,19 +258,15 @@ The last command must be executed as a privileged user if
 you are not currently using a virtualenv.
 
 
-After installation, add `django_celery_beat` to Django settings file and migrate.
+After installation, add ``django_celery_beat`` to Django settings file::
 
-settings.py
-```
-INSTALLED_APPS = [
-    ...,
-    'django_celery_beat',
-]
-```
+    INSTALLED_APPS = [
+        ...,
+        'django_celery_beat',
+    ]
 
-```
-python manage.py migrate django_celery_beat
-```
+    python manage.py migrate django_celery_beat
+
 
 Using the development version
 -----------------------------
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -26,7 +26,6 @@ from .models import (
     CrontabSchedule, IntervalSchedule,
     SolarSchedule, ClockedSchedule
 )
-from .utils import make_aware
 from .clockedschedule import clocked
 
 try:
@@ -136,7 +135,9 @@ class ModelEntry(ScheduleEntry):
 
         # CAUTION: make_aware assumes settings.TIME_ZONE for naive datetimes,
         # while maybe_make_aware assumes utc for naive datetimes
-        return self.schedule.is_due(maybe_make_aware(self.last_run_at).astimezone(self.app.timezone))
+        tz = self.app.timezone
+        last_run_at_in_tz = maybe_make_aware(self.last_run_at).astimezone(tz)
+        return self.schedule.is_due(last_run_at_in_tz)
 
     def _default_now(self):
         # The PyTZ datetime must be localised for the Django-Celery-Beat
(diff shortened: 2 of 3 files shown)


marky1991:
Doing that now. I think it's a bug in that test, but I'll fix it either way.

marky1991:
I found that the previously-failing test was failing based on the value of os.environ["TZ"], which I assume is why I couldn't reproduce it locally. Once I set TZ to the value of settings.TIME_ZONE, as django would,  I discovered that the tests were currently wrongly sending datetime.datetime.now(), i.e. now in local-time as last_run_at. This is not correct and only works if the local timezone is UTC (since then, the two are the same thing) This happens because celery's maybe_make_aware function, when given a naive datetime, calls to_utc on it, which merely adds UTC as the tzinfo and assumes that the time as given was in utc. Therefore, I changed the tests to pass utcnow instead of datetime.datetime.now. 

As part of this, I changed _default_now in this flow to instead use utcnow instead of datetime.now() and simply use maybe_make_aware. Currently, celery's maybe_make_aware and beat's make_aware handle naive datetimes differently, maybe_make_aware assuming utc and make_aware assuming localtime. I haven't changed this but simply stopped using make_aware in this flow. I checked other calls to make_aware and they look correct.

Let me know if there are any questions or if I should do something differently.

marky1991:
I finally got flake8 passing but the latest build failed to download pypy and thus didn't pass that one. (it passed every other time it ran though) Should I do something about this?

liquidpele:
I restarted the travisCI build so we'll see if it passes this time.

liquidpele:
It passed, but it looks like you have a conflict with master now...

marky1991:
I rebased with master again, so hopefully that fixes it.

liquidpele:
Looks like github still sees a conflict.  You'll want to merge upstream/master into your own branch and fix the conflict.   

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/merging-an-upstream-repository-into-your-fork

marky1991:
That's odd, it says there's no conflict for me:
"This branch has no conflicts with the base branch
Only those with write access to this repository can merge pull requests."

git pull upstream master
De https://github.com/celery/django-celery-beat
 * branch            master     -> FETCH_HEAD
Already up-to-date.

(The rebase version, git pull --rebase upstream master, makes me apply the same commits over and over again for some reason, I'm not sure why.)

Sorry for my git ignorance, I normally use hg.

Issue #310, opened 2020-02-16
troubleshooting a database scheduler which does nothing at all


superandrew:
### Summary:

Include a *brief* descrioption of the problem here, and fill out the version info below.  

* Celery Version: 
4.4.0
* Celery-Beat Version: 
1.4,1.5,1.6
### Exact steps to reproduce the issue:
1. Setup celery by reading the docs
2. setup some periodic task using cron, clocked schedule, periodi task
3. launch a beat and a worker instance

### Detailed information

I setup celery on heroku by using a dino for worker and one for beat (and one for web, obviously). I know that beat and worker are connected to redis and to postgres for task results.

Every periodic task I run from django admin by selecting a task and “run selected task” gets executed.

However, it is about two days that I’m trying to figure out why there isn’t a way for beat/worker to find that I scheduled a task to execute every 10 seconds, or using a cron (even restarting beat and remot doesn’t change it).

I’m kind of desperate, and my next move would be to give redbeat a try.

Any help on how to how to troubleshoot this particular problem would be greatly appreciated. I suspect the problem is in the `is_due` method. I am using UTC (in celery and django), all cron are UTC based. All I see in the beat log is “writing entries..” every on and then.

I’ve tried changing celery version from 4.3 to 4.4 and django celery beat from 1.4.0 to 1.5.0 to 1.6.0
Ideas?

superandrew:
it seems that beats stops processing data when there are dynamically created periodic tasks that in some ways interfere. Printing logs on the is_due function to debug reveals that only the `next` task is inspected, so if that task isn’t due any of the other tasks are inspected (for example my cron task). I that correct?

superandrew:
In case it helps someone who's having or will have a similar trouble as ours: to recreate this issue, it is possible to create a task as simple as:

```
@app.task(bind=True)
def test(self, arg):
    # f = open("testfile.txt", "w+")
    # f.write(arg)
    # f.close()
    print(kwargs.get("notification_id"))
```

then, in django admin, enter the task editing and put something in the extra args field. Or, viceversa, the task could be 

```
@app.task(bind=True)
def test(self, **kwargs):
    # f = open("testfile.txt", "w+")
    # f.write(arg)
    # f.close()
    print(kwargs.get("notification_id"))
```

[...]

PR #341, opened 2020-05-14
fix: fix clocked schedule

yywing:
Clocked schedule is stateless.
It can run once because of PeriodicTask.one_off.


django_celery_beat/models.py
```diff
@@ -186,12 +186,6 @@ class ClockedSchedule(models.Model):
         verbose_name=_('Clock Time'),
         help_text=_('Run the task at clocked time'),
     )
-    enabled = models.BooleanField(
```
auvipy, line 189: this is a breaking change

yywing: > can you explain more of your reason for this approach?

I maybe handle it before this weekends(timezone: Beijing). too busy now.

auvipy: we can introduce swapalbe/ abstract base model as Django-activity-stream did

yywing:
I will work on this PR.
And I maybe pull a new feature for mult time schedule.

yywing:
图片

yywing:
1. `is _due` return `None` will make schedule stuck. I dont know why and I canot find answer on celery document. But it works.
2. Clocked schedule changed clocked schedule model, and it will make schedule reload. As picture I upload.  
    1) I create many clocked schedule which work on the same time.
    2) first clocked schedule run ，and change model，and schedule reload.
    3) reload wiled call `is_due` and all clocked schedule set to false, but it olny called one
    4) schedule change and all schedule reload, all scheled set false at 2),so never called.

So only two task run.

yywing:
I change clocked model, remove enabled field, use `PeriodicTask.one_off`.

django_celery_beat/models.py
yywing: Or we can keep it but it always True.

superandrew:
I am looking forward for this PR, @yywing it is great that you have been able to track, understand fix one of the most hideous and awkward bugs of this extension.

superandrew:
just to add some hype: pip install git+https://github.com/yywing/django-celery-beat/@f2bdc5214b3c70ee3805654855ab08e55452df69
fixes the problem reported on #310

superandrew:
hi, any news on when this PR is going to be released?

auvipy:
there are some concerns regarding backward compat

superandrew:
too bad. My experience with my use case and django_celery_beat was that everything worked for days before stop working randomly (then I understood it was not literally randomly but there was a reason), and this fix definitely fixed everything. So I really hope this could go in production as soon as possible. I am not confident with celery or this package, but if I can help with some real world testing I'd be more than happy to do so.

auvipy:
please do some more real-world testing and report back

django_celery_beat/models.py
auvipy: what about default=False?

superandrew:
my production environment started using this package about 10 days ago.

We were able to run hundreds of task, and the scheduled tasks never stopped once.

Before this fix, after a few days, some condition would make celery beat block on one task and stop process the tasks all together.

I hope this helps. Please let me know if there is anything specific I could test.

django_celery_beat/models.py
yywing: But it act as always True.

sfdye:
Is it possible to release this fix please?

auvipy:
Next week Hopefully

adl-asi:
@auvipy Will this be merged soon? Trying to use clocked schedule in a production setting and running into issues

adl-asi:
Just checking in again. Is there an eta for when this fix will be released?

auvipy:
hope full this week. ping me again

saethor:
Any updates about this?

belongwqz:
still waiting

Lotram:
@auvipy  any news?

revoteon:
@auvipy Any updates?

ghazi-git:
I really wish I'd seen this PR earlier. I started using clocked schedule last month. Like everyone else, things started working fine and then at some point, the clocked tasks stopped being processed. What I found out is that whenever there is one task where `PeriodicTask.enabled` is `True` but `ClockedSchedule.enabled` is `False`, all tasks after it are never executed. So, the question became how did I end up in a situation where the task `PeriodicTask.enabled` is `True` but `ClockedSchedule.enabled` is `False`.

Later I was able to reproduce it consistently by scheduling a task, then stopping celery beat until the scheduled time of the task has passed. When celery beat is brought back up after that, the scheduled task, that should have executed when celery beat was down, is picked up and gets executed. So, `ClockedSchedule.enabled` is set to `False`, but then `PeriodicTask.enabled` is never changed to `False`, which results in the next tasks not executed. In production, that might be translated to "a clocked task needs to be scheduled to execute when there is some downtime/maintenance for the issue to start happening"

Anyway, right before deciding to switch to one-off tasks using crontab schedule, I saw that there is a new version and when checking the commits , I found this PR. Now, that `ClockedSchedule.enabled` is removed, I installed the new version and retested the issue and it is no longer happening. So, thank you @yywing for fixing this and @auvipy for releasing the fix.

sfdye:
Looks like `2.1.0` was released yesterday
https://pypi.org/project/django-celery-beat/#history

PR #346, opened 2020-06-03
Default timezone of CrontabSchedule from 'CELERY_TIMEZONE' setting.

### LINKED ISSUE #345: the default timezone of CrontabSchedule should not be a hardcode
### Summary:

the default timezone of CrontabSchedule current is 'UTC', can it keep the same with CELERY_TIMEZONE dynamically.  

* Celery Version: 4.4.3
* Celery-Beat Version: 2.0.0

1. open `xxx/admin/django_celery_beat/crontabschedule/add/ `  to add a crontabschedule
2. the Cron Timezone is UTC

### Detailed information

when is write `CELERY_TIMEZONE = 'Asia/Shanghai'` in the settings.py of my projects，can the timezone of CrontabSchedule is 'Asia/Shanghai' but not 'UTC'

mondeja:
After this pull, the default timezone of `CrontabSchedule` model is setted to the value defined in `CELERY_TIMEZONE` setting. If is not defined or is not a valid timezone, `"UTC"` is setted instead.

Closes #345


django_celery_beat/models.py
```diff
@@ -45,6 +45,17 @@ def cronexp(field):
     """Representation of cron expression."""
     return field and str(field).replace(' ', '') or '*'
 
+def crontab_schedule_celery_timezone():
```
auvipy, line 48: can you test this with unit test?

mondeja: Test added. Also, I've fixed the double space on `help_text` parameter of `timezone` `CrontabSchedule` model field taking advantage of the new migration.

belongwqz: A hardcode CELERY_TIMEZONE still here, when I `app.config_from_object('django.conf:settings', namespace='CELERY_OTHER')`, the setting name will be `CELERY_OTHER_TIMEZONE` but not `CELERY_TIMEZONE`.

auvipy: can you comment on Belongwqz comment

mondeja: Okay, seems that this do the trick:

```python
from celery import current_app

CELERY_TIMEZONE = getattr(settings, '%s_TIMEZONE' % current_app.namespace)
```

Can you review it again?

belongwqz: Yes, perfect!

django_celery_beat/migrations/0013_auto_20200603_1340.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.0.6 on 2020-06-03 13:40
```
auvipy, line 1: can you delete and regen this?

PR #347, opened 2020-06-09
Add SolarSchedule event choices i18n support

mondeja:
`SolarSchedule` model event choices translations are not recognized by `python manage.py makemessages`. To fix this I've hardcoded solar schedules readable representations for `SolarSchedule` model event choices.

- Solar schedules are keep synchronized with `celery.schedules.solar._all_events` tuple by a unit test.
- Translations for these new values have been added to `es` and `fr` locales.
- A new migration have been added to update `SolarSchedule` model event choices.
- Minor changes in Changelog to prevent `apicheck` error message: `Changelog:11:Field list ends without a blank line; unexpected unindent.`.


django_celery_beat/migrations/0014_auto_20200609_1048.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.6 on 2020-06-09 10:48
```
auvipy, line 1: this migration needs to be newly generated after pulling from master

auvipy (changes requested):

auvipy:
@arnau126 @asfaltboy mind to review this patch?

django_celery_beat/migrations/0014_auto_20200609_1048.py
mondeja: Do you mind if I wait for the full review of the patch?

asfaltboy (approved):
Seems legit, let's merge it :shipit: !

arnau126 (approved):

django_celery_beat/migrations/0014_auto_20200609_1048.py
auvipy: https://github.com/celery/django-celery-beat/blob/master/django_celery_beat/migrations/0014_remove_clockedschedule_enabled.py this is the last working migration which serial number seems same to this, thats why I was suggesting the remove and regenration :D

auvipy (changes requested):

django_celery_beat/migrations/0014_auto_20200609_1048.py
mondeja: Now it's pointing to `0014_remove_clockedschedule_enabled`. Let me know if any problem persists.

django_celery_beat/models.py
```diff
@@ -38,7 +38,17 @@
     (MICROSECONDS, _('Microsecond')),
 )
 
-SOLAR_SCHEDULES = [(x, _(x)) for x in sorted(schedules.solar._all_events)]
+SOLAR_SCHEDULES = [
```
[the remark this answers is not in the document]
mondeja: [Seeing the blame of the file](https://github.com/celery/django-celery-beat/blame/master/django_celery_beat/models.py) I can see that nothing related has changed and the global variable only is used in that file.

auvipy (approved):

Issue #150, opened 2018-06-27
Running one-off task stops all periodical tasks


putsi:
I'm running the development version which is installed via pip and noticed an issue related to the new one-off feature.

If there are n amount of tasks that are run periodically (e.g. once a minute) and the user runs some other task with the one-off attribute, the periodical tasks are not run anymore after the one-off task stops. 

Everything seems normal when inspected from the web interface (oneoff-task is disabled and the others are enabled) and from logs (scheduler keeps checking for tasks but doesn't simply notice/send the periodicals).
When the celery worker is restarted, the periodical tasks start running normally.

I tried to debug what is causing this but could not find any specific function.

hartwork:
Which version of django-celery-beat is affected?

vbabiy:
@putsi I have seen this as well. It is caused by 

```
    def is_due(self, last_run_at):
        if not self.enabled:
            return schedstate(is_due=False, next=None)
        rem_delta = self.remaining_estimate(None)
        remaining_s = max(rem_delta.total_seconds(), 0)
        if remaining_s == 0:
            if self.model:
                self.model.enabled = False
                self.model.save()
            return schedstate(is_due=True, next=None)
        return schedstate(is_due=False, next=remaining_s)
```

Not handling one of task correctly, and disabling the clocked schedule but not the one of task. Once in the this state the task is never due and blocks all other tasks during the tick.

I think #341 will fix this.

@auvipy I am seeing this on 2.0.0

auvipy:
> @putsi I have seen this as well. It is caused by
> 
> ```
>     def is_due(self, last_run_at):
>         if not self.enabled:
>             return schedstate(is_due=False, next=None)
>         rem_delta = self.remaining_estimate(None)
>         remaining_s = max(rem_delta.total_seconds(), 0)
>         if remaining_s == 0:
>             if self.model:
>                 self.model.enabled = False
>                 self.model.save()
>             return schedstate(is_due=True, next=None)
>         return schedstate(is_due=False, next=remaining_s)
> ```
> 
> Not handling one of task correctly, and disabling the clocked schedule but not the one of task. Once in the this state the task is never due and blocks all other tasks during the tick.
> 
> I think #341 will fix this.
> 
> @auvipy I am seeing this on 2.0.0

i request you to review that PR and change that in a nonbreakable way

yywing:
`schedstate(is_due=False, next=None)` will make it stuck.

hartwork:
[...]

PR #350, opened 2020-06-13
If the clocked schedule is tied to a one off task, don't disable it. …

vbabiy:
As far as I can tell there is no reason to disable the clock, sine that the task will get disabled on the next run.


vbabiy:
This should fix #150

yywing:
I dont think so. I update my pr.

auvipy:
> I dont think so. I update my pr.

which PR, please?

auvipy (changes requested):
rebase please

auvipy:
this change is not relevant anymore

PR #352, opened 2020-06-13
Keep meta data around task, even after disabled.

vbabiy:
- added `total_run_count` to the admin
- don't clear out last_run_at, on disabling a task.
- don't clear out total_run_count on one of tasks.

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -574,8 +574,6 @@ class PeriodicTask(models.Model):
         self.routing_key = self.routing_key or None
         self.queue = self.queue or None
         self.headers = self.headers or None
-        if not self.enabled:
-            self.last_run_at = None
         self._clean_expires()
         self.validate_unique()
         super(PeriodicTask, self).save(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


django_celery_beat/models.py
```diff
@@ -574,8 +574,6 @@ def save(self, *args, **kwargs):
         self.routing_key = self.routing_key or None
         self.queue = self.queue or None
         self.headers = self.headers or None
-        if not self.enabled:
```
auvipy, line 577: does this create any regreassion?

django_celery_beat/schedulers.py
```diff
@@ -125,7 +125,6 @@ def is_due(self):
         if self.model.one_off and self.model.enabled \
                 and self.model.total_run_count > 0:
             self.model.enabled = False
-            self.model.total_run_count = 0  # Reset
```
auvipy, line 128: is it backward-incompatible?

auvipy (changes requested):
please come with an up to date solution

Issue #360, opened 2020-09-10
Signature in periodic task


SunnyCapt:
### Summary:

The link parameter (task callback) is not supported also chain, group etc tasks are not supported in periodic tasks.

* Celery Version: latest
* Celery-Beat Version: latest

### Detailed information

Periodic chain tasks are too hard to create (you need to implement new celery task which creates a task chain)

PR #361, opened 2020-09-10
Impl: signature in periodic task

SunnyCapt:
changelog:

* added support for task signature (by dill lib)
* added support for task callback signature (by dill lib)

close: #360


auvipy (changes requested):
check the unit test failures please

auvipy:
can you rebase please?

django_celery_beat/utils.py
```diff
@@ -11,6 +17,68 @@
 # see Issue #222
 now_localtime = getattr(timezone, 'template_localtime', timezone.localtime)
 
+logger = get_logger(__name__)
+
+
+def generate_keys(
```
auvipy, line 23: I am not sure why we need rsa key?

SunnyCapt: it's need to sign serialized objects so as not to execute any python code that may appear in the database (execute only signed serilized code which can be signed and added only from main proc)

SunnyCapt: https://davidhamann.de/2020/04/05/exploiting-python-pickle/

auvipy: why noy json serializer? it's default in celery

auvipy: JSON serializer should be the default task serializer in the integration packages as well

SunnyCapt:
@auvipy, could you please help fix the tests?

django_celery_beat/utils.py
SunnyCapt: okay, I'll rewrite

auvipy: dill is ok for pickle, but pickle is not the default serializer. so rewriting this with json would be great

PR #365, opened 2020-09-27
fix "ModuleNotFoundError: No module named 'celery.utils.encoding'"

torchingloom:


auvipy (changes requested):
can you also check the test failures?

thedrow:
> can you also check the test failures?

The test failures are unrelated.
We need to invest some work here since we've done some changes in Celery 5 which break the test suite like introducing `pytest-celery`.

thedrow (approved):
LGTM

auvipy (approved):

mfonville:
Please also publish it as an updated release, so that we can all install this fix directly with `pip` :-)

auvipy:
let me check all the related open issues before a pypi release

PR #378, opened 2020-11-29
Fix breakage caused by change in django-timezone-field

### LINKED ISSUE #377: Latest Version of django-timezone-field breaks django-celery-beat
Hello,

I wrote a more detailed issue at this link:
https://github.com/pydanny/cookiecutter-django/issues/2950

Generally, the owner of django-time-zone updated today (Nov. 28th, 2020) their package to version 4.1.1, which broke the migration 0013_auto_20200609_0727.py.

The reason this happened is two fold however, since you guys have https://github.com/celery/django-celery-beat/blob/master/requirements/default.txt saying: "django-timezone-field>=4.0,<5.0", so when he/she updated their pypi, django-celery-beat installation from pip pulled the latest version of django-timezone-field instead of the stable version, which breaks this package since it now uses the latest 4.1.1 version. 

Mandating django-timezone-field==4.0 in my requirements.txt file fixes the issue.

Thanks,
Omar

davidrothera:
In https://github.com/mfogel/django-timezone-field/commit/0c6a6bae8e9c36c3d236e47eff136b085a4639df the `CHOICES` property was removed, it was replaced with two new fields `default_tzs` and `default_choices`. The latter of the two is the same choices as before and so we just use that now instead.

Fixes #377

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -57,7 +57,8 @@ def crontab_schedule_celery_timezone():
     except AttributeError:
         return 'UTC'
     return CELERY_TIMEZONE if CELERY_TIMEZONE in [
-        choice[0].zone for choice in timezone_field.TimeZoneField.CHOICES
+        choice[0].zone for choice in timezone_field.
+        TimeZoneField.default_choices
     ] else 'UTC'
 
 
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,3 +1,3 @@
 celery>=4.4,<6.0
-django-timezone-field>=4.0,<5.0
+django-timezone-field>=4.1.0,<5.0
 python-crontab>=2.3.4
--- a/t/unit/test_models.py
+++ b/t/unit/test_models.py
@@ -60,7 +60,8 @@ class MigrationTests(TestCase):
 
 
 class CrontabScheduleTestCase(TestCase):
-    FIRST_VALID_TIMEZONE = timezone_field.TimeZoneField.CHOICES[0][0].zone
+    FIRST_VALID_TIMEZONE = timezone_field.\
+        TimeZoneField.default_choices[0][0].zone
 
     def test_default_timezone_without_settings_config(self):
         assert crontab_schedule_celery_timezone() == "UTC"


arnau126:
With this change, `django-celery-beat` will no longer work with `django-timezone-field==4.0`.

I see two options:
1. Modify the minimum required version of `django-timezone-field` in [requirements](https://github.com/celery/django-celery-beat/blob/master/requirements/default.txt): `django-timezone-field>=4.1,<5.0`
2. Make the code work with both versions: modify  `django_celery_beat/models.py` so it uses `CHOICES` or `default_choices` depending on the `django-timezone-field` version.

davidrothera:
Great point @arnau126, I had meant to pin the requirements but forgot in the initial PR

davidrothera:
Another option would be to parse and gate by using the `timezone_field.__version__` field with something like the `packaging.version.parse` method but it would require another third-party dependency (one already used by setup tools though so its more than likely already installed)

ohld (approved):

django_celery_beat/models.py
```diff
@@ -57,7 +57,8 @@ def crontab_schedule_celery_timezone():
     except AttributeError:
         return 'UTC'
     return CELERY_TIMEZONE if CELERY_TIMEZONE in [
-        choice[0].zone for choice in timezone_field.TimeZoneField.CHOICES
+        choice[0].zone for choice in timezone_field.
```
auvipy, line 60: I guess this also need change in model migrations?

auvipy (changes requested):

django_celery_beat/models.py
davidrothera: @auvipy I don't think any other changes will be needed, the tuples outputted by `TimeZoneField.CHOICES` and `default_choices` are the same and so the data should be the same.

davidrothera:
```
(venv) ~/D/django-celery-beat ❯❯❯ python3.9 manage.py makemigrations                                                                         master
No changes detected
```

auvipy (approved):

tasiotas:
hi, Im not quite knowlegable with github, PR and merging.
I am using:
django-celery-beat         2.1.0
django-timezone-field      4.1.1

and getting this error:
django_celery_beat\models.py", line 60, in crontab_schedule_celery_timezone
    choice[0].zone for choice in timezone_field.TimeZoneField.CHOICES
AttributeError: type object 'TimeZoneField' has no attribute 'CHOICES'

What can I do about it?

arnau126:
This PR has fixed your error.
It is merged, but not released.
Until then (django-celery-beat 2.1.1 is released) downgrade django-timezone-field to 4.0.

tasiotas:
> This PR has fixed your error.
> It is merged, but not released.
> Until then (django-celery-beat 2.1.1 is released) downgrade django-timezone-field to 4.0.

Thank you!

Issue #322, opened 2020-03-14
The task celery.backend_cleanup will delete my tasks added before if have the same crontab time


linxi-1214:
I set my task with crontab <0, 4, *, *, *>, then restart celery. 
I find my task was deleted, then I debug the code and I find a snippet executed in `django-celery-beat/models.py`:

```
@classmethod
    def from_schedule(cls, schedule):
        spec = {'minute': schedule._orig_minute,
                'hour': schedule._orig_hour,
                'day_of_week': schedule._orig_day_of_week,
                'day_of_month': schedule._orig_day_of_month,
                'month_of_year': schedule._orig_month_of_year,
                'timezone': schedule.tz
                }
        try:
            return cls.objects.get(**spec)
        except cls.DoesNotExist:
            return cls(**spec)
        except MultipleObjectsReturned:
            cls.objects.filter(**spec).delete()
            return cls(**spec)
```
I wonder why delete the crontab if `MultipleObjectsReturned` error occurred even if the crontab is bound with other task?

mvikharev:
+1

izimobil:
This is a major issue indeed, just came across it recently.
Probably for historic reasons and because there is no unique constraint, schedules tables can contain duplicates.
So, schedules should not be deleted as they can be linked to other tasks, IMO the correct code should be:

```python
@classmethod
    def from_schedule(cls, schedule):
        spec = {'minute': schedule._orig_minute,
                'hour': schedule._orig_hour,
                'day_of_week': schedule._orig_day_of_week,
                'day_of_month': schedule._orig_day_of_month,
                'month_of_year': schedule._orig_month_of_year,
                'timezone': schedule.tz
                }
        try:
            return cls.objects.get(**spec)
        except cls.DoesNotExist:
            return cls(**spec)
        except MultipleObjectsReturned:
            cls.objects.filter(**spec).first()
```
I'll send a pull request ASAP.

auvipy:
looking forward to it

hartwork:
Can probably be closed now that #389 is merged…

auvipy:
> Can probably be closed now that #389 is merged…

shouldn't we track https://github.com/celery/django-celery-beat/pull/269 for future?

hartwork:
> > Can probably be closed now that #389 is merged…
> 
> shouldn't we track #269 for future?

I don't have "need" for that myself right now but if someone wants to team up on the topic, I can think my way back into it. As you like.

auvipy:
[...]

PR #389, opened 2021-03-02
Do not blindly delete duplicate schedules

izimobil:
Because there are no unique constraints on schedule models (except for
SolarSchedule), schedules tables can contain duplicates, those
duplicate schedules should not be blindly deleted when encountered
because they can be linked to existing tasks and would cause the tasks
to be also deleted (on delete cascade).

Instead we now just return the first duplicate found, just to avoid
creating further duplicates.

This fixes issue #322.

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -122,8 +122,9 @@ class SolarSchedule(models.Model):
         except cls.DoesNotExist:
             return cls(**spec)
         except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+            # unique_together constraint should not permit reaching this code,
+            # but just in case, we return the first schedule found
+            return cls.objects.filter(**spec).first()
 
     def __str__(self):
         return '{0} ({1}, {2})'.format(
@@ -183,8 +184,7 @@ class IntervalSchedule(models.Model):
         except cls.DoesNotExist:
             return cls(every=every, period=period)
         except MultipleObjectsReturned:
-            cls.objects.filter(every=every, period=period).delete()
-            return cls(every=every, period=period)
+            return cls.objects.filter(every=every, period=period).first()
 
     def __str__(self):
         readable_period = None
@@ -236,8 +236,7 @@ class ClockedSchedule(models.Model):
         except cls.DoesNotExist:
             return cls(**spec)
         except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+            return cls.objects.filter(**spec).first()
 
 
 class CrontabSchedule(models.Model):
@@ -350,8 +349,7 @@ class CrontabSchedule(models.Model):
         except cls.DoesNotExist:
             return cls(**spec)
         except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+            return cls.objects.filter(**spec).first()
 
 
 class PeriodicTasks(models.Model):
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -117,14 +117,13 @@ class SolarSchedule(models.Model):
         spec = {'event': schedule.event,
                 'latitude': schedule.lat,
                 'longitude': schedule.lon}
+
+        # we do not check for MultipleObjectsReturned exception here because
+        # the unique_together constraint safely prevents from duplicates
         try:
             return cls.objects.get(**spec)
         except cls.DoesNotExist:
             return cls(**spec)
-        except MultipleObjectsReturned:
-            # unique_together constraint should not permit reaching this code,
-            # but just in case, we return the first schedule found
-            return cls.objects.filter(**spec).first()
 
     def __str__(self):
         return '{0} ({1}, {2})'.format(
--- a/t/unit/test_models.py
+++ b/t/unit/test_models.py
@@ -80,17 +80,25 @@ class CrontabScheduleTestCase(TestCase):
 
     def test_duplicate_schedules(self):
         # See: https://github.com/celery/django-celery-beat/issues/322
+        kwargs = {
+            "minute": "*",
+            "hour": "4",
+            "day_of_week": "*",
+            "day_of_month": "*",
+            "month_of_year": "*",
+            "day_of_week": "*",
+        }
         # create 2 duplicates schedules
-        sched1 = CrontabSchedule.objects.create(hour="4")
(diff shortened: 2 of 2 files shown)


auvipy:
can you check if this PR https://github.com/celery/django-celery-beat/pull/269/files is somewhat related?

izimobil:
> can you check if this PR https://github.com/celery/django-celery-beat/pull/269/files is somewhat related?

It is related in the sense that it would prevent future duplicates, the idea of adding unique constraints is good, but there should be a warning in the changelog because it's not fully backwards compatible, existing code may try to create duplicate schedules (which currently works) and get a `IntegrityError` exception with the constraints added.

Furthermore, because existing databases can already contain duplicate schedules, the `from_schedule` methods should not be modified.

IMO you can merge my PR first (the bug is really critical in the sense that it deletes existing tasks !), release a bugfix version, and add constraints on a future major version.

(If the PR is reworked to only add constraints, the tests on my PR will have to be removed.)

auvipy:
@arnau126 can you check this please

auvipy (commented):
hope return cls.objects.filter(**spec).first() won't create any regression as bug it is bug fix

izimobil:
> hope return cls.objects.filter(**spec).first() won't create any regression as bug it is bug fix

Frankly, it can't be worse than it is now, I got tasks that disappeared on my production environment and it's really difficult to repair when you have hundreds of tasks.

auvipy:
just waiting for another eye for cross-checking. will merge soon after that

t/unit/test_models.py
```diff
@@ -74,6 +78,20 @@ def test_default_timezone_without_settings_config(self):
     def test_default_timezone_with_settings_config(self):
         assert crontab_schedule_celery_timezone() == self.FIRST_VALID_TIMEZONE
 
+    def test_duplicate_schedules(self):
+        # See: https://github.com/celery/django-celery-beat/issues/322
+        # create 2 duplicates schedules
+        sched1 = CrontabSchedule.objects.create(hour="4")
+        CrontabSchedule.objects.create(hour="4")
+        self.assertEqual(CrontabSchedule.objects.count(), 2)
```
hartwork, line 86: Just an idea:
I think adding `.filter(hour="4")` here and in line 93 would be a cheap way to make this test more robust to any potential cross-test effects. It would help fight future flakiness. What do you think?
PS: This applies to the other two tests as well.

t/unit/test_models.py
```diff
@@ -99,3 +117,43 @@ def test_celery_solar_schedules_included_as_event_choices(self):
 
         for event_choice in event_choices_values:
             assert event_choice in schedules.solar._all_events
+
+    def test_duplicate_schedules(self):
+        # Duplicates cannot be tested for solar schedules because of the
+        # unique constraints in the SolarSchedule model
+        pass
```
hartwork, line 124: I'm not a big fan of adding dead code, to be honest. There are other things that we cannot test for either, right? Does this code really add value?

django_celery_beat/models.py
```diff
@@ -122,8 +122,9 @@ def from_schedule(cls, schedule):
         except cls.DoesNotExist:
             return cls(**spec)
         except MultipleObjectsReturned:
-            cls.objects.filter(**spec).delete()
-            return cls(**spec)
+            # unique_together constraint should not permit reaching this code,
+            # but just in case, we return the first schedule found
+            return cls.objects.filter(**spec).first()
```
hartwork, line 127: In the interest of avoiding dead code, why not drop these lines altogether?

hartwork (commented):
Hi!

This improves over the existing code for sure. :+1:
What I am wondering is:
- Why are we using `cls()` over `cls.objects.create()` so that those functions sometimes return model instances that *are* persisted already and sometimes return instances that *are not* yet persisted in the database. Is that a bug or a feature?
- It seems like we effectively have a tailor-made `get_or_create` here.  Why not use what Django offers and simplify most of this code away for good?
- The added test cases are effectively 3 times the same test.  Personally, I would use [`parameterized.expand`](https://pypi.org/project/parameterized/) and a single parametrized test to avoid duplication, but it's just my two cents and I'm aware there there are multiple schools on that subject.

Best, Sebastian

django_celery_beat/models.py
izimobil: Given the constraint is here from the very beginning (https://github.com/celery/django-celery-beat/commit/ec46367c38a63a01e3e9ea6992534bc914650746) we can indeed drop the MultipleObjectsReturned block

t/unit/test_models.py
izimobil: See next comment

hartwork: I checked all. Which one?

izimobil: Given the constraint is here from the very beginning (ec46367) we can indeed drop the MultipleObjectsReturned block (as well as the test method of course)

hartwork: So the test would be dropped as well, excellent. :+1:

izimobil:
Hi,
> 
>     * Why are we using `cls()` over `cls.objects.create()` so that those functions sometimes return model instances that _are_ persisted already and sometimes return instances that _are not_ yet persisted in the database. Is that a bug or a feature?
> 
>     * It seems like we effectively have a tailor-made `get_or_create` here.  Why not use what Django offers and simplify most of this code away for good?

I asked myself the same question regarding why the schedule isn't saved, but keep in mind I don't want to break anything with this bug fix PR.  Improving the code should be the role of another PR IMHO.

> 
>     * The added test cases are effectively 3 times the same test.  Personally, I would use [`parameterized.expand`](https://pypi.org/project/parameterized/) and a single parametrized test to avoid duplication, but it's just my two cents and I'm aware there there are multiple schools on that subject.

I won't take the responsability of adding yet another requirement just for that, instead I've factorized the code in a mixin class.

Thanks for your review, regards.

hartwork (approved):
@izimobil thanks for taking this further! :+1: :pray:

hartwork:
Travis CI taking ages to start builds as usual… :man_shrugging:

izimobil:
@auvipy I think you can safely merge this one, I can't believe no one else came across the dramatic impacts of previous code !
Cheers.

auvipy:
caught by cold flu.  but reviewing and merging it tomorrow for sure

auvipy (approved):

hartwork:
> caught by cold flu. but reviewing and merging it tomorrow for sure

All the best!

PR #410, opened 2021-05-26
Fix MySQL max length issue.

jeffkit:
Fix the same issue in django_celery_result project.


auvipy (commented):
is this change really needed?

jeffkit:
> is this change really needed?

Yes, this issue is also exist in the earlier version of django_celery_result  project. Check the [readme](https://github.com/celery/django-celery-results#issues-with-mysql) for detail.

we use utf8-mb4 as our mysql default chartset encoding, when migrating django_celery_beat, it fail with an error:
`django.db.utils.OperationalError: (1071, 'Specified key was too long; max key length is 767 bytes')`

auvipy:
hopefully the change is not going to break existing codes

auvipy:
can you check the comments of this PR and improve this one https://github.com/celery/django-celery-beat/pull/202/files?

PR #464, opened 2021-12-03
Fix ClockedSchedule and PeriodicTasks showing UTC time when Time Zone is enabled.

oreaba:
**This contribution fixes 2 serious issues when displaying timing on Django Admin:**

1- When time zone is used, celery beat ClockedSchedule model shows two different times which causes discrepancies.
For example, When you enter the django admin > Clocked > list display item, you will find the time is right (w.r.t your TZ), like: Dec. 3, 2021, 6:11 p.m.
However, when you click on this item to open the detailed view it will show the UTC time of the clock you entered reesulting in discrepancies, like: 2021-12-03 16:11:57+00:00

2- When time zone is used, celery beat PeriodicTasks model shows the the name of the task and the datetime of the task in UTC not in the local time zone chosen, which causes great confusion in the periodic tasks.
For example, the PeriodicTasks list will have a task like this: "Send Email: 2021-12-03 16:20:57+00:00" but as a matter of fact this should be 6:20 PM not 4:20 PM as shown in the list, although I have CELERY_ENABLE_UTC = False and CELERY_TIMEZONE             = 'Africa/Cairo'

This fix solves both problems and shows the correct time with respect to time zone.


django_celery_beat/models.py
```diff
@@ -220,7 +221,8 @@ class Meta:
         ordering = ['clocked_time']
 
     def __str__(self):
-        return '{}'.format(self.clocked_time)
+        return localtime(self.clocked_time).strftime("%d %B %Y %I:%M %p")
```
auvipy, line 224: can unit test added for this change?

auvipy (changes requested):

django_celery_beat/models.py
lvelvee: I think `return '{}'.format(make_aware(self.clocked_time))` is better.

oreaba: Yes @lvelvee I believe so. 
I will make another pull request with this change.

oreaba: Yes @lvelvee I believe so.
I have done another commit with this change after fetching the new base code.

auvipy:
can you add small unit test for it?

auvipy (approved):

oreaba:
> 
@auvipy 
Yes sure 
I have written it.
I will commit it today.

auvipy:
ok please come with it today please

oreaba:
> ok please come with it today please

Done #485

PR #478, opened 2021-12-15
Adding name col to CrontabSchedule model

trianglesis:
Having a lot of Cron time schedule items is frustrating when they have no verbose and short names, for example.

image

I'm trying to add a name to each - so it will be more visible, what task run on which time, actually.

image
image

With this change, it becomes more visible for the human eye what schedule is used for each task.
It's also better visible when you want to switch schedules, but the dropdown list is too big.

image

In addition: for years of usage - I have collected dozens of unused schedules and it starting to become a nightmare when you want to clean it up.


django_celery_beat/migrations/0016_crontabschedule_name.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.1.7 on 2021-12-15 13:06
```
auvipy, line 1: can you use django 3.2.x to regenerate this

auvipy (commented):
also pull from master please

django_celery_beat/migrations/0016_crontabschedule_name.py
trianglesis: Re-created with 3.2

auvipy:
@lvelvee kindly have a review when have time

django_celery_beat/migrations/0016_crontabschedule_name.py
lvelvee: @trianglesis  It doesn't seem necessary to do the `if` twice. I'll checkout your branch in my projects.

lvelvee: looks good

django_celery_beat/models.py
```diff
@@ -315,11 +315,16 @@ class Meta:
                     'day_of_week', 'hour', 'minute', 'timezone']
 
     def __str__(self):
-        return '{0} {1} {2} {3} {4} {5} (m/h/dM/MY/d) {6}'.format(
-            self.name if self.name else '',
+
+        name = None
+        if self.name:
+            name = '[{}]'.format(self.name)
```
trianglesis, line 321: Here I wanted to achieve [Name in brackets], but leave an empty '' if there is no name. 
Making if and if - makes it more visible on code, but It could be changed back to 1st version.
Thanks!

django_celery_beat/models.py
```diff
@@ -299,6 +299,13 @@ class CrontabSchedule(models.Model):
             'Timezone to Run the Cron Schedule on. Default is UTC.'),
     )
 
+    name = models.CharField(
```
auvipy, line 302: would also like to have unit tests for new proposed changes

auvipy (changes requested):

django_celery_beat/models.py
trianglesis: I can't run any test from test_schedulers.py with the current setup. 
Will try to figure out what am I doing wrong, and possibly add some tests.

django_celery_beat/admin.py
```diff
@@ -116,9 +116,9 @@ class PeriodicTaskAdmin(admin.ModelAdmin):
     model = PeriodicTask
     celery_app = current_app
     date_hierarchy = 'start_time'
-    list_display = ('__str__', 'enabled', 'interval', 'start_time',
```
auvipy, line 119: why str is removed?

auvipy (changes requested):
unit tests for the changes to avoid possible regression?

trianglesis:
I'll try to add tests as soon as I figure out how to execute them all. Unfortunately, I can't get to work few tests related to model testing, which I wanted to use as a draft\example to test my changes.

trianglesis:
Currently 
- testing the creation of Crontab with valid names or an empty string.
- testing that PeriodicTask.name has no extra values in it and if PeriodicTask.interval and PeriodicTask.crontab are in a human-readable format (as it was before in name).

auvipy (changes requested):
can you please rebase the conflicts please?

auvipy (changes requested):
can you please revisit this PR and see if this change is still a good addition for this package? I tried to fix the conflicts but there are too many

trianglesis:
This is so outdated now from the idea I had before. 

This PR is not vital now, since I achieved the desired view using the custom `admin view`.

trianglesis:
Closing it for good.

auvipy:
thanks

PR #489, opened 2021-12-23
Update test.yml

saschahofmann:
Use test.txt requirement


saschahofmann:
@auvipy I assume this is what you wanted to use? Am I getting this right, right now the tests arent run in CI?

auvipy (approved):
I am adding this but the CI needs more work

PR #501, opened 2022-01-24
Change last_run_at=None when using disable tasks admin action

Kirade:
PeriodicTask model's save method implements that when disabling task, it changes its `last_run_at` field to None.

```
# models.py
if not self.enabled:
    self.last_run_at = None
```

But `disable_tasks` in PeriodicTaskAdmin action. does not change periodic task's `last_run_at` attribute.

```
# admin.py
def disable_tasks(self, request, queryset):
    rows_updated = queryset.update(enabled=False)
    PeriodicTasks.update_changed()
    self._message_user_about_update(request, rows_updated, 'disabled')
disable_tasks.short_description = _('Disable selected tasks')
```

Due to this behavior, Even if we disable some tasks using action, tasks runs immediatey as soon as task enabled.

To fix this issue, I suggest change `last_run_at` attribute to None when disable_tasks.

Thanks!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -37,6 +37,7 @@ Grégoire Cachet <gregoire@audacy.fr>
 Hari <haridara@gmail.com>
 Idan Zalzberg <idanzalz@gmail.com>
 Ionel Maries Cristian <ionel.mc@gmail.com>
+Jaeyoung Heo <jay.jaeyoung@gmail.com>
 Jannis Leidel <jannis@leidel.info>
 Jason Baker <amnorvend@gmail.com>
 Jay States <jstates@based.ca>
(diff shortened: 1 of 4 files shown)


auvipy (changes requested):
thanks for the PR can you check the lint failure please? I will do some testing & thorough review before merge

Kirade:
Yes, I'll check out and request you soon. Thanks!

auvipy (changes requested):
can you also include release note entry please?

Kirade:
> can you also include release note entry please?

Yup, I guess you mean Changelog entry. right? 😉

auvipy (approved):

auvipy:
thanks. feel free to send any follow up PR, if any regression happen

PR #503, opened 2022-01-24
Remove duplicated translation of "Enabled"

Kirade:
while locale message compile using `python manage.py compilemessages`
Command complains about french locale.

Screen Shot 2022-01-24 at 8 45 20 PM

I found it `msgid="Enabled"` has defined twice!
So I removed old one. and now compile works well.

Thanks

### DIFF
--- a/django_celery_beat/locale/fr/LC_MESSAGES/django.po
+++ b/django_celery_beat/locale/fr/LC_MESSAGES/django.po
@@ -229,10 +229,6 @@ msgstr "Horaire"
 msgid "Run the task at clocked time"
 msgstr "Démarre la tâche à l'horaire définie"
 
-#: django_celery_beat/models.py:216 django_celery_beat/models.py:516
-msgid "Enabled"
-msgstr "Activée"
-
 #: django_celery_beat/models.py:217 django_celery_beat/models.py:517
 msgid "Set to False to disable the schedule"
 msgstr "Mettre à Faux pour désactiver la planification"


django_celery_beat/locale/fr/LC_MESSAGES/django.po
```diff
@@ -229,10 +229,6 @@ msgstr "Horaire"
 msgid "Run the task at clocked time"
 msgstr "Démarre la tâche à l'horaire définie"
 
-#: django_celery_beat/models.py:216 django_celery_beat/models.py:516
```
auvipy, line 232: can you point to another one of which it is duplicate?

auvipy (changes requested):

django_celery_beat/locale/fr/LC_MESSAGES/django.po
Kirade: Sure.

https://github.com/celery/django-celery-beat/blob/6172ba4b484cb4173ec7a28a31573ef26877cd8c/django_celery_beat/locale/fr/LC_MESSAGES/django.po#L233

https://github.com/celery/django-celery-beat/blob/6172ba4b484cb4173ec7a28a31573ef26877cd8c/django_celery_beat/locale/fr/LC_MESSAGES/django.po#L471

auvipy (approved):
your work is right based on https://github.com/celery/django-celery-beat/blob/master/django_celery_beat/models.py#L507

PR #520, opened 2022-03-03
Replace pytz with zoneinfo

ResleyR:
Replace pytz with zoneinfo, which is part of Python 3.9+ and the default timezone implementation for Django 4.x onwards

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -46,7 +46,7 @@ Important Warning about Time Zones
 
 
 
-.. note:: 
+.. note::
    This will reset the state as if the periodic tasks have never run before.
 
 
@@ -95,7 +95,7 @@ create the interval object:
 .. code-block:: Python
 
         >>> from django_celery_beat.models import PeriodicTask, IntervalSchedule
-        
+
         # executes every 10 seconds.
         >>> schedule, created = IntervalSchedule.objects.get_or_create(
         ...     every=10,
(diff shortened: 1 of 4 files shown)


requirements/runtime.txt
```diff
@@ -1 +1,4 @@
 Django>=2.2,<4.1
+# zoneinfo is preset in Python 3.9 onwards
+# Django 4.x onwards uses zoneinfo by default but not the tzdata
+backports.zoneinfo[tzdata]==0.2.1; python_version < '3.9'
```
auvipy, line 4: can we unpin the strict versioning?

README.rst
```diff
@@ -181,7 +181,7 @@ of a ``30 * * * *`` (execute every 30 minutes) crontab entry you specify:
         ...     day_of_week='*',
         ...     day_of_month='*',
         ...     month_of_year='*',
-        ...     timezone=pytz.timezone('Canada/Pacific')
+        ...     timezone=zoneinfo.ZoneInfo('Canada/Pacific')
```
auvipy, line 184: will this work in older django versions?

ResleyR: I was hoping the CI would run all the tests under tox...         

I'm getting myself familiar with the project and I feel there are more changes needed. ZoneInfo should work but I will test against 2.2.x, 3.2.x and 4.x just to be sure

ResleyR: I've been trying to run tox and it fails in v2.2.1 as well (timeout)               
In master it's far worse with errors (since it installs [django-timezone-field 5.0](https://pypi.org/project/django-timezone-field/)

In this branch (after few more changes) Django 2.2 fails with the same timeout... so does Django 4.0....   
but Django 3.2 has a conflict installing Django

> The conflict is caused by:
>     The user requested django<4.0 and >=3.2.12
>     The user requested django<4.1 and >=4.0.2

not sure which dependency is causing this

auvipy:
i think we should postpone the migration until we drop support of django 2.2.

auvipy (changes requested):
it is fixed in new release by another PR>

Issue #532, opened 2022-04-11
Support for compose environment variables and interpolation


diegocastrum:
### Summary:

To avoid port collision when working on `django-celery-beat` in some devel environments it would be nice if the `docker-compose` makes use of env variables defined in a `.env` file, for more information about how docker-compose loads these variables take a look of the docker-compose [documentation](https://docs.docker.com/compose/environment-variables/).

### Detailed information

The idea is just to make use of string interpolation from `docker-compose.yml` taking the current values that everybody uses as defaults, When the project runs, if a `.env` file is found with the right env var defined, docker-compose will use it instead the default one. This approach shouldn't any current development setup

PR #533, opened 2022-04-11
Added support for compose env vars (#532)

diegocastrum:
Added `.env` to `.gitignore`.

Added a `.env.sample` that will be used by anyone he need it. Just copy it as `.env` in your local devel env and use the supported) variables you need.

`docker-compose.yml` updated in order to support string interpolation when using variables defined in a `.env` file.

Fixed a typo in `issue_template.md`.


.gitignore
```diff
@@ -33,3 +33,4 @@ coverage.xml
 .eggs/
 .python-version
 venv
+.env
```
auvipy:
```suggestion
.env

```

auvipy (approved):

PR #563, opened 2022-07-08
Remove superseded ExtendedQuerySet as it's functionality is built into Django since 1.7

### LINKED ISSUE #562: ExtendedQuerySet.update_or_create violates signature of parent method
### Summary:

`ExtendedQuerySet.update_or_create` violates signature of parent method and `ExtendedQuerySet`s functionality is part of Django since years anyways.

* Celery Version: 
* Celery-Beat Version: 

1. Call PeriodicTask.objects.update_or_create
2. Expect the same signature as with all other QuerySets
3. Fail

### Detailed information

`ExtendedQuerySet.update_or_create` extends from Django's default `django.db.models.query.QuerySet` but changes the signature of the method. 
See https://github.com/django/django/blob/608ab043f75f1f9c094de57d2fd678f522bb8243/django/db/models/query.py#L949 the parent method returns a tuple of `object, created: bool` whereas the derived one only returns the `object` and there's no way to figure out if the task has been changed or not.

tbh I don't know if this is a chicken/egg problem. `update_or_create` was introduced to Django in 1.7 on 2014/09/02 and the initial commit to `django-celery-beat` was on 2016/08/04 but named `initial import` so there might have been some code before that.

Imho the signature should be aligned with the parent method or `ExtendedQuerySet` should be completely removed as Django provides the same functionality now for years.

Let me know what you prefer and I'd be more than happy to open a pull request.

lociii:
Fixes #562

`ExtendedQuerySet` provided a custom implementation of `update_or_create` before the release of Django 1.7. Since then Django added it to the base `QuerySet`.  
Only difference is that `update_or_create` of [Django's](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#update-or-create) `QuerySet` returns the object and creation state whereas the custom implementation of `django-celery-beat` only returned the object. Therefore `ModelEntry.from_entry` had to be adjusted.

As a cleanup, I removed all managers and introduced a single queryset for `PeriodicTask`.


auvipy (changes requested):
builds are failing. also some additional tests would be great

lociii:
> builds are failing. also some additional tests would be great

What kind of tests would be like to see?
`PeriodicTaskQuerySet` now only has one method left. 

> builds are failing. also some additional tests would be great

Builds are fixed now as far as I can influence them.
What kind of tests are you looking for? I can test `PeriodicTaskQuerySet` but that seems pretty pointless to me.

auvipy:
can you add a little test for the queryset?

954-Ivory:
> can you add a little test for the queryset?

This PR just use django native model function, to replace custom function.
It should not to add additional test code.

I think this PR is better in django version>1.7:
Because it will call native function, the native function will call with `transaction.atomic`, it's safer.

But I don't know does django has `update_or_create` in version <1.7.
If there has not `update_or_create`, some user use django-celery-beat with django<1.7 will got the error.

auvipy:
this will be django 2.2+ only

auvipy (approved):

PR #564, opened 2022-07-11
Set options['expire_seconds'] from model.expire_seconds

bbkgh:


auvipy (changes requested):
whats problem will this change solve? can you add test to verify this?

auvipy:
@954-Ivory can you verify this patch?

954-Ivory:
**Do not** approved this PR.

https://github.com/celery/django-celery-beat/blob/626c2a06331a039250ea4026812bf1ede26907cc/django_celery_beat/models.py#L599-L600

https://github.com/celery/django-celery-beat/blob/626c2a06331a039250ea4026812bf1ede26907cc/django_celery_beat/models.py#L592-L596

 `expires` or `expire_seconds` only one can be set.

954-Ivory:
wait

954-Ivory:
I went to test just now, `expire_seconds` seems doesn't work in master branch.

auvipy (changes requested):
need regression test to verify this

Issue #572, opened 2022-08-21
Signals can not connect to `OneToOneField`.


954-Ivory:
Because django let `sender = self.__class__`:
https://github.com/django/django/blob/e9fd2b572410b1236da0d3d0933014138d89f44e/django/db/models/base.py#L860

If you inheritance from `PeriodicTask`(same as `OneToOneField`), and overwrite `save()` in your field, like this:

``` python3
class YourPeriodicTask(PeriodicTask):
    # ...
    def save(self, *args, **kwargs):
        # ...
        super().save(*args, **kwargs)
```

The `pre_save.sender` will be `YourPeriodicTask`, not `PeriodicTask`.
So, the signal will not be receive.

auvipy:
is it bug or what you are suggesting?

954-Ivory:
> is it bug or what you are suggesting?

Yeah, I think it's a bug.
The PR comming soon~

PR #573, opened 2022-08-21
Fixed signals can not connect to OneToOneField (#572)

954-Ivory:
Incidentally, the file storage of the signal has been changed.


django_celery_beat/signals.py
```diff
@@ -0,0 +1,41 @@
+def signals_connect():
+    from django.db.models import signals, OneToOneRel
```
auvipy, line 2: OneToOneRel is a private API as far as I know, can you provide a testt case of examples of what you were actually trying to acheive?

auvipy (changes requested):

django_celery_beat/signals.py
954-Ivory: Test case:

```python3
# admin.py
class MyPeriodicTaskInline(admin.StackedInline):
    model = MyPeriodicTask
    # ...
    # I want to render a `StackedInline` of `MyPeriodicTask` in `OtherModelAdmin`.

class OtherModelAdmin(admin.ModelAdmin):
    model = OtherModel
    inlines = (
        MyPeriodicTaskInline,
    )
```
```python3
# models.py
class MyPeriodicTask(PeriodicTask):
    other_model = models.ForeignKey(
        'OtherModel',
        null=True,
        on_delete=models.CASCADE
    )
    def save(self, *args, **kwargs):
        self.kwargs = json.dumps('a_field_key',self.other_model.a_field)
        super().save(*args, **kwargs)
        # The bug is in it: 
        # This `save()` will call the `save_base()`.
        # But in `save_base()`, it will set `pre_save.sender = MyPeriodicTask`, not `PeriodicTask`.
        # So, the `PeriodicTask` can not receive signal with
        # [signals.pre_delete.connect(PeriodicTasks.changed, sender=PeriodicTask)](https://github.com/celery/django-celery-beat/blob/d4b97af62889527582194f3932e013ad972a88b3/django_celery_beat/models.py#L625)
        # [signals.pre_save.connect(PeriodicTasks.changed, sender=PeriodicTask)](https://github.com/celery/django-celery-beat/blob/d4b97af62889527582194f3932e013ad972a88b3/django_celery_beat/models.py#L625)
```
Even without not overwrite `save()`, the `pre_save.sender` still be `MyPeriodicTask.__class__`.
Because in `MyPeriodicTask`, the `self` is `MyPeriodicTask`.
Amd Django use [self.__class__](https://github.com/django/django/blob/e9fd2b572410b1236da0d3d0933014138d89f44e/django/db/models/base.py#L860) to set `sender`.
`delete()` is the same, look [this](https://github.com/django/django/blob/e9fd2b572410b1236da0d3d0933014138d89f44e/django/db/models/deletion.py#L433).

954-Ivory: We have two way fix it.
1. Don't use signal, call `PeriodicTasks.changed()` in `PeriodicTask.save()` and `PeriodicTask.delete()`.
2. Like this PR, use reverse related to get all `OneToOneRel`, then get thier `__class__`, connect to the signal.

auvipy: i would prefer first approach

django_celery_beat/models.py
```diff
@@ -585,6 +585,11 @@ def save(self, *args, **kwargs):
         self._clean_expires()
         self.validate_unique()
         super().save(*args, **kwargs)
+        PeriodicTasks.changed(self)
+
+    def delete(self, *args, **kwargs):
+        super(PeriodicTask, self).delete(*args, **kwargs)
```
auvipy, line 591:
```suggestion
        super().delete(*args, **kwargs)
```

auvipy (changes requested):

django_celery_beat/signals.py
auvipy: can you also add the tests as well

954-Ivory: > can you also add the tests as well

Yep, but I got the problem like this:
https://stackoverflow.com/questions/502916/django-how-to-create-a-model-dynamically-just-for-testing

auvipy: is that related to this patch?

auvipy:
looking for now-outdated files... none found
  pickling environment... done
  checking consistency... done
  
  ==============================
   Undocumented Autodoc Modules 
  ==============================
  
  py
  --
  
  - django_celery_beat.signals

django_celery_beat/signals.py
954-Ivory: > 



> is that related to this patch?

Yep, I solved it.
Because I need to create the model in testing.

954-Ivory:
> looking for now-outdated files... none found pickling environment... done checking consistency... done
> 
> # ==============================
> Undocumented Autodoc Modules
> ## py
> * django_celery_beat.signals

I consolidated some files about `Signals` before (follow Django best practices).
Does it conflict about CI/CD?

auvipy:
the doc gen expects some documentation about the newly created module

954-Ivory:
> the doc gen expects some documentation about the newly created module

Right, I will add the docs.

auvipy:
some minor Error: ERROR:   flake8: commands failed
Error: ERROR:   pydocstyle: commands failed need to be fixed

954-Ivory:
\

> some minor Error: ERROR: flake8: commands failed Error: ERROR: pydocstyle: commands failed need to be fixed

It seems like some code specification problem.
I have tried to format them.
The `migrations` file is be generated, I didn't format it.
How to ignore code check of the `migrations` file?

auvipy (approved):

auvipy:
lets try this and report if anything goes wrong

auvipy:
can you help reviewing https://github.com/celery/django-celery-beat/pull/563?

Issue #604, opened 2022-11-08
Celery Beat Crashing at the End of Daylight Savings


polarmt:
### Summary:
* Celery Version: 4.4.7
* Celery-Beat Version: 2.2.0 

### Exact steps to reproduce the issue:
1. Set `CELERY_TIMEZONE` equal to `America/Los_Angeles`
2. Run the Celery beat such that an `IntervalSchedule` task runs between 1 AM and 2 PM on the day that daylight savings ends. In our case, this was 2022-11-06.
3. Celery beat will crash with the following traceback:

```
[INFO] [2022-11-06 01:18:52,155] [18813] [schedulers.schedule:356] - DatabaseScheduler: Schedule changed.
Process Beat:
Traceback (most recent call last):
  File "/path/to/virtualenv/lib/python3.7/site-packages/billiard/process.py", line 327, in _bootstrap
    self.run()
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 707, in run
    self.service.start(embedded_process=True)
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 631, in start
    interval = self.scheduler.tick()
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 329, in tick
    self.populate_heap()
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/beat.py", line 303, in populate_heap
    is_due, next_call_delay = entry.is_due()
  File "/path/to/virtualenv/lib/python3.7/site-packages/django_celery_beat/schedulers.py", line 135, in is_due
    return self.schedule.is_due(last_run_at_in_tz)
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/schedules.py", line 171, in is_due
    rem_delta = self.remaining_estimate(last_run_at)
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/schedules.py", line 137, in remaining_estimate
    self.maybe_make_aware(self.now()), self.relative,
  File "/path/to/virtualenv/lib/python3.7/site-packages/celery/schedules.py", line 76, in now
    return (self.nowfun or self.app.now)()
  File "/path/to/virtualenv/lib/python3.7/site-packages/django_celery_beat/models.py", line 175, in <lambda>
    nowfun=lambda: make_aware(now())
  File "/path/to/virtualenv/lib/python3.7/site-packages/django_celery_beat/utils.py", line 27, in make_aware
[...]

PR #605, opened 2022-11-11
Fix Issue #604: Celery Beat Crashing at the End of Daylight Savings

polarmt:
## Overview
This is my suggested fix for https://github.com/celery/django-celery-beat/issues/604.

## Testing
All testing was done on the following versions:
django: 3.2.15
celery: 5.2.7
django-celery-beat: 2.4.0

`USE_TZ` is not set.

### Reproducing the Issue
In the Django shell (`python manage.py shell`):
```python
from datetime import datetime, timedelta
from celery.utils import time
from django_celery_beat.utils import make_aware
from pytz import timezone

dt  = datetime(2022, 11, 6, 1, 34, 0)
now = make_aware(dt)
```

Without the changes in this PR, this should throw an `AmbiguousTimeError`.

### Behavior of `remaining` During Transition (Without ...)
We can run the following script to test how the Celery beat will calculate the time remaining during the transition:

```
from datetime import datetime, timedelta
from celery.utils import time
from django_celery_beat.utils import make_aware
from pytz import timezone

dt  = datetime(2022, 11, 6, 1, 34, 0)
now = make_aware(dt)

tz = timezone('America/Los_Angeles')
start = datetime(2022, 11, 6, 1, 15, 0)
start = tz.localize(start, is_dst=True)
ends_in = timedelta(minutes=80)

print(time.remaining(time.maybe_make_aware(start), ends_in, time.maybe_make_aware(now)))
```

`now` here would be `1:34 AM PST` or `1:34:00-8:00` while `start` would be `1:15 AM PDT` or `1:15:00-7:00`. The correct return value of `time.remaining(...)` here would be 1 minute since 1 hour 20 minutes from `start` would be `1:35:00-8:00`. The problem is that `time.remaining(...)` returns 1 hour 1 minute because `time.remaining` will convert `1:15:00-7:00` to `1:15-8:00`.

###  Behavior of `remaining` During Transition (With ...)
We can use the same script as the above subsection after making the changes in . This will cause `time.remaining(...)` to return 1 minute instead of 1 hour 1 minute, which is the correct value.


django_celery_beat/utils.py
```diff
@@ -24,7 +26,15 @@ def make_aware(value):
     else:
         # naive datetimes are assumed to be in local timezone.
         if timezone.is_naive(value):
-            value = timezone.make_aware(value, timezone.get_default_timezone())
+            tm_isdst = time.localtime().tm_isdst
+
+            # tm_isdst may return -1 if it cannot be determined
```
auvipy, line 31: I am not sure about this part, can any other way we can check it?

auvipy (changes requested):
beside it will be very useful to have proper unit tests for this change

django_celery_beat/utils.py
polarmt: I referenced these posts [1](https://stackoverflow.com/questions/2881025/python-daylight-savings-time) (top answer only) and [2](https://stackoverflow.com/questions/39610892/in-python-how-to-check-if-a-given-time-is-in-daylight-saving-time-period). There are other suggested solutions online. I can summarize the problems with these solutions.

### Use of `localize`
[Source 1](https://gist.github.com/dpapathanasiou/09bd2885813038d7d3eb)
[Source 2](https://stackoverflow.com/questions/2881025/python-daylight-savings-time)
[Source 3](https://stackoverflow.com/questions/31146092/how-to-determine-if-a-timezone-specific-date-in-the-past-is-daylight-saving-or-n)

The root cause was inside `localize`. Using the `localize` to fix the problem with `localize` sounds counterintuitive. For instance, link 1 mentions the following in the final solution:

```
Once you've **localized** a datetime in fact you can simply check it's .dst() method to see if it's zero.
```

### `time.daylight`
[Source](https://www.geeksforgeeks.org/python-time-daylight-function/)

This solution might work, but the underlying code is just to do a similar check to what I do here. This code is cleaner, and I can change it if you would like.

### Calculations
[Source](https://stackoverflow.com/questions/38722792/check-if-date-is-in-daylight-savings-time-for-timezone-without-pytz)

One way is to calculate the timezone based on the current time and a previous time, but I think that this solution is long and too ad-hoc when we can achieve this with a few lines of code.

### Conclusion
I think that `time.daylight` or `time.localtime`+`tm_isdst` is the best way to go for this. If you believe in the other solutions, please let me know.

polarmt: Sounds good. Let me know when you've checked.

cclauss:
Now that tests are passing on Django v5 this all becomes more difficult because:
* Support for `pytz` timezones is removed.
* The `is_dst` argument is removed from:
    * `QuerySet.datetimes()`
    * `django.utils.timezone.make_aware()`

https://docs.djangoproject.com/en/5.0/releases/5.0/#features-removed-in-5-0

cclauss:
@polarmt Please rebase to resolve the git conflicts.

cclauss:
@polarmt Please rebase to resolve the git conflicts.

cclauss:
@polarmt Tests are ~50/50 pass fail and in Django v5 [_the default value of the USE_TZ setting is changed from False to True_](https://docs.djangoproject.com/en/5.0/releases/5.0/#features-removed-in-5-0).

Now that we have release v2.6.0 that supports Django v5, how should we proceed with this effort?

auvipy (changes requested):
we need to fix the conflict and make the ci green as the related pr seems to be passing in celery

auvipy:
@alirafiei75 sorry to ping you again here. since you are doing a lot recently, can you please check this PR and related celery issue, that if it is needed any more? if yes, if possible send a new pr with fixing the old codes?

alirafiei75:
> @alirafiei75 sorry to ping you again here. since you are doing a lot recently, can you please check this PR and related celery issue, that if it is needed any more? if yes, if possible send a new pr with fixing the old codes?

Not a problem at all. Happy to help. I will check it in the upcoming week.

eht16:
@alirafiei75 any chance you might find time to update the PR to get it merged?

I was hit by this issue, or maybe https://github.com/celery/celery/issues/6438 last weekend at the DST change.
The monitoring alerted because a Beat task wasn't executed for one hour at the DST change and caused me to get up in the night to check the alert :cry:.

PR #613, opened 2022-12-16
Clarify month range

rapto:
Current example shows 0 as a valid value for month. Valid range is 1-12, as the validator correctly detects.


auvipy (approved):

auvipy:
sorry need to revert this as the changes proposed need additional schema migrations so a new PPR with the additional migration is requested

PR #615, opened 2022-12-17
Clarify month range

rapto:
Current example shows 0 as a valid value for month. Valid range is 1-12, as the validator correctly detects.

It includes the corresponding migration.


auvipy (approved):

auvipy:
thanks a lot

PR #618, opened 2022-12-20
Add Python 3.11 support

gopackgo90:


gopackgo90:
FYI, after I closed this PR I made an issue with the maintainer of `python-crontab` and they just published a wheel for the latest release of the library, 2.6.0, https://pypi.org/project/python-crontab/2.6.0/#files, so the issue with `python-crontab` is likely not an issue anymore, but `ephem` suffered from the same bdist_wheel problem.

PR #621, opened 2022-12-22
Update PeriodicTaskAdmin list display

khanh96le:
PR for issue #620

Issue #612, opened 2022-12-15
More friendly crontab in Django Admin


khanh96le:
### Summary:

In Django admin, I see that it would be more user-friendly if we can replace the crontab format by natural language
Something similar like this: https://crontab.guru

* Celery Version: 4.4.7
* Celery-Beat Version: 2.2.0

### Exact steps to reproduce the issue:
1. Go to Django Admin
2. Access `Periodic Tasks` > `Periodic Tasks`
3. Look at the `Periodic task` column

### Detailed information

Current
![Screen Shot 2022-12-15 at 12 43 31](https://user-images.githubusercontent.com/19431276/207782038-21c1917f-678e-4957-b56d-275af79f1ab3.png)

Expected
![Screen Shot 2022-12-15 at 12 45 00](https://user-images.githubusercontent.com/19431276/207782149-d96d656f-54ec-4d2a-bb58-6690cd3ded6a.png)

truthdoug:
A first step would be to use something like [cron-descriptor](https://github.com/Salamek/cron-descriptor) to show human readable descriptions of the schedules.

auvipy:
contributions are welcome

truthdoug:
Like this contribution? ☝️

auvipy:
yes left my review comments

PR #622, opened 2022-12-22
Adding human readable descriptions of crontab schedules

truthdoug:
Partially addresses #612

## What I did
* incorporated [cron-descriptor](https://github.com/Salamek/cron-descriptor) to provide more readable descriptions of crontab schedules
* added these descriptions on the PeriodicTask detail view on and the Crontab list view
* updated help text for Crontab to help me remember how the numbers translate to days of the week
* modest updates to locales

## What is missing
* testing of non-English locales
* translation of new help text for locales other than English, Spanish, and German 

## Possible problems
* The human readable text of _all_ crontabs is loaded into the context of the PeriodicTask detail view. This could be a problem for sites that have a very large number of crontab schedules.


django_celery_beat/models.py
```diff
@@ -276,8 +277,8 @@ class CrontabSchedule(models.Model):
         max_length=64, default='*',
         verbose_name=_('Day(s) Of The Week'),
         help_text=_(
-            'Cron Days Of The Week to Run. Use "*" for "all". '
-            '(Example: "0,5")'),
+            'Cron Days Of The Week to Run. Use "*" for "all", Sunday '
```
auvipy, line 280: this changes might new a new schema migration. can you try makemigrations and see anything generate and commit that as well?

auvipy (changes requested):

django_celery_beat/models.py
truthdoug: Good catch. Committed and squashed into my original commit for improving the helptext.

django_celery_beat/admin.py
```diff
@@ -138,9 +138,25 @@ class PeriodicTaskAdmin(admin.ModelAdmin):
...
 
+    def crontab_translation(self, obj):
+        return obj.crontab.human_readable
+
+    change_form_template = 'admin/djcelery/change_periodictask_form.html'
+
+    def changeform_view(self, request, object_id=None, form_url='',
+                        extra_context=None):
+        extra_context = extra_context or {}
+        crontabs = CrontabSchedule.objects.all()
```
auvipy, line 152: instead of calling all, can we reduce the number?

django_celery_beat/templates/admin/djcelery/change_periodictask_form.html
```diff
@@ -0,0 +1,22 @@
...
+    <script id="periodic-task-overrides">
+        const readableCrontabs = JSON.parse(django.jQuery("#readable-crontabs").text());
+        var originalCrontabHelp = "{{ adminform.form.fields.crontab.help_text }}";
+
+        var updateCrontabHelp = function(additional) {
+            django.jQuery(".field-crontab .help").html(
+                `${originalCrontabHelp}<br/>Translation: ${additional}`);
+        };
+
+        django.jQuery(".field-crontab_translation").hide()
```
auvipy, line 16: is this django.jQuery being used from django admin itself? is it possible to switch to plain vanilla JS here? ES 6 without any dependencies?

auvipy (changes requested):

django_celery_beat/templates/admin/djcelery/change_periodictask_form.html
truthdoug: From [django documentation](https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#jquery):
> To avoid conflicts with user-supplied scripts or libraries, Django’s jQuery (version 3.6.0) is namespaced as django.jQuery. 

Since this namespaced jQuery comes with Django admin, this seems like the best choice for using jQuery. It's probably possible to use ES 6 w/o dependencies but jQuery is not an extra dependency, so I think using this is ok.

django_celery_beat/admin.py
truthdoug: Hmmm, I'm not sure how we'd decide which to include. My goal here is to provide natural language translations for all available crontabs.

The best argument for reducing the number is possible slow response times when loading the page. But, based on other sites I've used that load a lot of data into javascript objects onto the DOM, my hunch is that there would have to be hundreds (if not thousands) of CrontabSchedule objects before it would impact performance of this page. Additionally, since this is an admin-only view, any hit to responsiveness would never be seen by end users.

Do you have thoughts on how we might choose which ones to reduce it to?

auvipy: can we show 20/25 per page?

auvipy (changes requested):

django_celery_beat/admin.py
truthdoug: It's not clear to me how paginating this would work. 

My goal with this change is to update the human readable description anytime the user changes the select list of defined crontabs. So, if I change from `5 * * * *` to `0 10 * * 1`, I want the description to change from "At 5 minutes past the hour" to "At 10:00 AM, only on Monday" to verify that the change is what I expect.

If the code only loads 20 per page, what happens if I choose a crontab that wasn't part of the 20 that were loaded?

This could be implemented as ajax/XHR so that the description is readable only on demand. Since django-celery-beat doesn't currently have any custom endpoints defined, this seems like a bigger change. Alternately, there could be a setting to disable this if a site has so many CrontabSchedules defined that loading this admin page is onerous.

Can you tell me more about your objection to having all defined on this page?

auvipy:
I am not sure why CI is failing. can you check please?

django_celery_beat/models.py
```diff
@@ -616,7 +626,7 @@ def schedule(self):
         if self.interval:
             return self.interval.schedule
         if self.crontab:
-            return self.crontab.schedule
+            return self.crontab
```
khanh96le, line 629: Some unit tests are failed because of this change.
Can you elaborate more on why we need this?

khanh96le: I recently closed PR #621 because I realized the changes are already covered in yours.
Hope this can give you an idea.

auvipy: I can merge part of your PR first @khanh96le

khanh96le: Sure @auvipy. I saw that you already reopened it

auvipy:
builds are now passing after merging your PR @khanh96le

django_celery_beat/models.py
truthdoug: I like the change you made in #621, @khanh96le! That's an improved list view!

auvipy (approved):

PR #627, opened 2023-01-23
Declare support for Django 4.2

nijel:
- add tests
- adjust requirements
- adjust classes


snmishra:
I'd suggest doing away with the upper limit on Django version. It's unnecessary and causes breakage for projects using this library.

auvipy (approved):

Issue #647, opened 2023-03-24
Exception in admin: invalid literal for int() with base 10: '1DAY'


nijel:
### Summary:

Include a *brief* description of the problem here, and fill out the version info below.  

* Celery Version: 5.2.7
* Celery-Beat Version: 2.5.0

### Exact steps to reproduce the issue:
1. Create task with cron schedule `0 2 * * monday (m/h/dM/MY/d) UTC` (created via Python `crontab(hour=2, minute=0, day_of_week="monday")`)
2. Open it in admin interface
3. Get a crash

### Detailed information

Please include more detailed information here, such as relevant information about your system setup, things you did to try and debug the problem, log messages, etc.  

```
Traceback (most recent call last):
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/cron_descriptor/ExpressionDescriptor.py", line 131, in get_full_description
    day_of_week_desc = self.get_day_of_week_description()
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/cron_descriptor/ExpressionDescriptor.py", line 342, in get_day_of_week_description
    return self.get_segment_description(
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/cron_descriptor/ExpressionDescriptor.py", line 467, in get_segment_description
    description = get_description_format(expression).format(get_single_item_description(expression))
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/cron_descriptor/ExpressionDescriptor.py", line 345, in <lambda>
    lambda s: get_day_name(s),
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/cron_descriptor/ExpressionDescriptor.py", line 315, in get_day_name
    return ExpressionDescriptor.number_to_day(int(exp))

During handling of the above exception (invalid literal for int() with base 10: '1DAY'), another exception occurred:
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/django/core/handlers/exception.py", line 56, in inner
    response = get_response(request)
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/weblate/weblate-env/lib/python3.9/site-packages/sentry_sdk/integrations/django/views.py", line 85, in sentry_wrapped_callback
    return callback(request, *args, **kwargs)
[...]

PR #648, opened 2023-03-27
Avoid crash when can not get human readable description

nijel:
Catch errors from cron-descriptor and display crontab as it is. Apparently there are crontabs which work in Celery, but cron-descriptor fails to format them.

Issue #647

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -8,7 +8,8 @@ from datetime import timedelta
 
 import timezone_field
 from celery import current_app, schedules
-from cron_descriptor import get_description
+from cron_descriptor import (FormatException, MissingFieldException,
+                             WrongArgumentException, get_description)
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


auvipy:
@Salamek can you have a look please?

Salamek:
@auvipy It is ok as a hotfix (to prevent a crash)... but i think you should create a valid CRON expression from `crontab` object  and pass it to cron_descriptor to get correctly displayed...

auvipy (approved):

schinckel:
For what it's worth, I was getting this error in the admin.

I was able to resolve it without new code by updating the crontab data to change `wednesday` to `wed` (for example).

auvipy:
> For what it's worth, I was getting this error in the admin.
> 
> I was able to resolve it without new code by updating the crontab data to change `wednesday` to `wed` (for example).

should we mention this somewhere n docs?

Issue #388, opened 2021-02-26
Using django-celery-beat， Scheduled tasks are triggered repeatedly


ugfly1210:
```
celery==4.4.6
django-celery-beat==2.1.0
Django==3.1.2
django-celery-results==2.0.0
```

Set a timing policy: execute the task at 00:10 every day task_A
look at the beat log shows that
```
[2021-02-26 00:10:00,052: INFO/MainProcess] Scheduler: Sending due task task_A
[2021-02-26 00:10:05,194: INFO/MainProcess] Scheduler: Sending due task task_A
```

celery log:
```
[2021-02-26 00:10:00,061: INFO/MainProcess] Received task: task_A[e30e7941-a7da-4a9f-90d9-b125f8e4c46]   expires:[2021-03-29 16:01:59.944841+00:00]
[2021-02-26 00:10:00,120: INFO/ForkPoolWorker-5] Task task_A[e30e7941-a7da-4a9f-90d9-fb125f8e4c46] succeeded in 0.05772967217490077s: True
[2021-02-26 00:10:05,233: INFO/MainProcess] Received task: task_A[c9e7ce40-664d-4bf8-a14d-84651a5885eb]   expires:[2021-03-29 16:02:05.186303+00:00]
[2021-02-26 00:10:05,298: INFO/ForkPoolWorker-5] Task task_A[c9e7ce40-664d-4bf8-a14d-84651a5885eb] succeeded in 0.06434300076216459s: True
```

other configuration parameters:
```
BROKER_BACKEND = 'redis'
CELERY_ACCEPT_CONTENT = ['application/json']
CELERY_RESULT_SERIALIZER = 'json'
CELERY_TASK_SERIALIZER = 'json'
CELERY_TASK_RESULT_EXPIRES = 60 * 60 * 24
CELERYD_CONCURRENCY = 4  
CELERYD_MAX_TASKS_PER_CHILD = 4  
CELERYD_PREFETCH_MULTIPLIER = 4
CELERYD_FORCE_EXECV = True
# BROKER_TRANSPORT_OPTIONS = {'visibility_timeout': 90}
CELERY_CREATE_MISSING_QUEUES = True
CELERY_IGNORE_RESULT = True
CELERY_SEND_EVENTS = False
CELERY_DISABLE_RATE_LIMITS = False
# CELERY_EVENT_QUEUE_EXPIRES = 60
CELERY_TIMEZONE = TIME_ZONE
CELERY_ENABLE_UTC = False
[...]

PR #660, opened 2023-07-10
Fix Issue #388: Celery Beat scheduled tasks may be executed repeatedly

BaiJiangJie:
## Describe

Fix the problem that Celery Beat scheduled tasks may be executed repeatedly

### Describe

If `self.schedule` is called, it may cause the data in the database to be retrieved before the `self.save()` method, instead of the temporarily set `last_run_at data`

If `self.schedule` is called here
Then it may cause the `last_run_at` of `self.schedule[name]` calling save to be the old data retrieved from the database
Instead of `last_run_at` temporarily set after task execution (set in `__next__()` method)
After the `max_interval` interval, the next task detection cycle will still execute the task again

### Demo

Task information:
   >beat config: max_interval = 60s
  
   >Task name: cap
   >Task execution cycle: Execute every 3 minutes
   >Task last execution time: 18:00

The first execution of the task: 18:03 (set `last_run_at` = 18:03 when executing, and it is in memory at this time)
                    
   >After the task execution is completed,
   >It is detected that sync is required, and `self.schedule` is called in `sync`,
   >If `schedule_changed()` is found to be `True` in `self.schedule`, `all_as_schedule()` needs to be called
   >At this point, the `last_run_at` of `self.schedule[name]` called in `sync` is 18:00
   >At this time, `self.save()` is performed in `self.sync()`

beat: Waking up 60s...
                    
The second execution of the task: 18:04 (because the `last_run_at` obtained is 18:00, `entry.is_due()` == `True`)

## Test
There are currently no test cases.

## Effect
After modifying the code, the problem of repeated execution no longer appeared in our project [JumpServer](https://github.com/jumpserver/jumpserver/pull/10936/files).

### DIFF
--- a/t/unit/test_schedulers.py
+++ b/t/unit/test_schedulers.py
@@ -444,6 +444,28 @@ class test_DatabaseScheduler(SchedulerCase):
         assert self.s.flushed == 1
         assert self.m2.name in self.s._dirty
 
+    def test_sync_not_saves_last_run_at_while_schedule_changed(self):
+        # Update e1 last_run_at and add to dirty
+        e1 = self.s.schedule[self.m2.name]
+        time.sleep(3)
+        e1.model.last_run_at = e1._default_now()
+        self.s._dirty.add(e1.model.name)
+
+        # Record e1 pre sync last_run_at
+        e1_pre_sync_last_run_at = e1.model.last_run_at
+
+        # Set schedule_changed() == True
+        self.s._last_timestamp = monotonic()
+        # Do sync
+        self.s.sync()
+
+        # Record e1 post sync last_run_at
+        e1_m = PeriodicTask.objects.get(pk=e1.model.pk)
+        e1_post_sync_last_run_at = e1_m.last_run_at
+
+        # Check
+        assert e1_pre_sync_last_run_at == e1_post_sync_last_run_at
+
     def test_sync_saves_last_run_at(self):
         e1 = self.s.schedule[self.m2.name]
         last_run = e1.last_run_at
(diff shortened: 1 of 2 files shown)


BaiJiangJie:
https://github.com/celery/django-celery-beat/issues/388

auvipy (commented):
you can try to add new tests for this. also if yyou can fix the failing test in this or another PR, it will be really helpful

BaiJiangJie:
你从代码逻辑上来看这样修改是对的吗？
如果没问题，我再考虑增加它的测试用例。

Is it right to modify it in this way from the logical point of view of code?

If there is no problem, I will consider adding its test cases.

auvipy:
I think you can proceed with tests

BaiJiangJie:
Okay, it might take a while.

t/unit/test_schedulers.py
```diff
@@ -444,6 +444,28 @@ def test_reserve(self):
...
+        # Do sync
+        self.s.sync()
+
+        # Record e1 post sync last_run_at
+        e1_m = PeriodicTask.objects.get(pk=e1.model.pk)
+        e1_post_sync_last_run_at = e1_m.last_run_at
+
+        # Check
+        assert e1_pre_sync_last_run_at == e1_post_sync_last_run_at
+
```
BaiJiangJie, line 468: image

BaiJiangJie: It is normal if self.schedule is changed to self._schedule.

image

BaiJiangJie:
The test case has been added, you can review it, thank you.

@auvipy

auvipy (approved):

PR #666, opened 2023-07-22
Django celery beat 1.0.1 that supports django4

RonShub:


auvipy (changes requested):
use the latest release please

Issue #691, opened 2023-10-25
Python 3.12


foarsitter:
Tested Python 3.12 in #630 but `case` isn't working since it relies on `nose`. 

There are plans to archive `case` in https://github.com/celery/case/issues/13 

A shortcut we can take is to move the `patching` fixture from `case` to this package. Thoughts?

PR #692, opened 2023-10-25
Replace case.patching fixture with mockeypatch + MagicMock

foarsitter:
The amount of failing test-cases put my on the wrong foot since patching was used in an autouse fixture.

This PR has a minimal implementation of the `patching` fixture of `case` and drops `case` as dependency. as required in #691


auvipy (approved):
thanks

PR #697, opened 2023-11-28
Allow IntervalSchedule to use a custom schedule type.

sdrabblescripta:


auvipy (changes requested):
we will need tests to verify the proposed changes

sdrabblescripta:
Of course, and I planned on adding them but I'm unable to run the current test suite so if you could provide advice on how to do that successfully (`tox -e py38` returns a bunch of errors) then I'll be happy to add new tests.

auvipy (changes requested):
can you please describe why we actually need this?

auvipy:
> Of course, and I planned on adding them but I'm unable to run the current test suite so if you could provide advice on how to do that successfully (`tox -e py38` returns a bunch of errors) then I'll be happy to add new tests.

can you also share the errors? running tox should be easy. you can use the python version install in your path initially

PR #707, opened 2023-12-21
Replace flake8 and isort with ruff

cclauss:
### https://docs.astral.sh/ruff


.pre-commit-config.yaml
```diff
@@ -2,22 +2,6 @@
 exclude: "migrations"
 
 repos:
-  - repo: https://github.com/asottile/pyupgrade
-    rev: v3.15.0
-    hooks:
-      - id: pyupgrade
-        args: ["--py37-plus"]
```
thedrow, line 9: Why remove pyupgrade?

cclauss: It is built into ruff... https://docs.astral.sh/ruff/rules/#pyupgrade-up

auvipy (commented):
I want to do a thorough review before merge, thanks for the PR

PR #705, opened 2023-12-19
Testing Django v5.0 on pypy-3.10-v7.3.14 passes

cclauss:
Identical to #699 but ___adds a single tox test that fails___: Django v5.0 on pypy-3.10-v7.3.14.
* #699
* https://github.com/celery/django-celery-beat/pull/699#issuecomment-1863399975

---
`tox_gh_actions` prefers the hyphen in `pypy-3.10` (not `pypy3.10`)
https://github.com/ymyzk/tox-gh-actions/blob/master/src/tox_gh_actions/plugin.py#L174

Fixes #708
* #708

PR #699, opened 2023-12-06
Upgrade GitHub Actions and PyPy 3.10 and Django 5.0

cclauss:
Upgrading our project from Django v4.2 to v5.0 ___downgraded___ our versions of django-celery-beat and django-timezone-field!

`Pipfile.lock`:
```diff
        "django-celery-beat": {
            [ ... ]
-           "version": "==2.5.0"
+           "version": "==2.1.0"
        },
```
% `python3.12 -m venv .venv`
% `source .venv/bin/activate`
% `pip install --upgrade pip`
% `pip install django==5.0 django-celery-beat==2.5.0` --> **_Failure_**
% `pip install django==5.0 django-celery-beat` --> Success but with
* `django-celery-beat` downgraded `2.5.0` --> `v2.1.0` and
* `django-timezone-field` downgraded from `v6.1.0`--> `v4.2.3`

Please release a new version to PyPI.
* #680
* #698

Issue #698, opened 2023-12-05
django 5.0


qengine13:
Installed django 5.0. and got the warning that celery beat 2.5.0 not compatible... Will there be an update?

aminvsf:
I was wondering the same thing. It's the only package among my other dependencies which is incompatible.

piotr-grzelka:
#681

tuky:
So it only needs a new release? Considering how simple the patch was, I would suggest to loosen the deps a little bit and at least support Django < 6 and not only Django < 5.1. Nobody can tell the future, but IMO package maintainers should focus on the minimum dependency versions and packages users on maximum dependency versions.

cclauss:
% `python3.12 -m venv .venv`
% `source .venv/bin/activate`
% `pip install --upgrade pip`
% `pip install django==5.0 django-celery-beat==2.5.0` --> **_Failure_**
% `pip install django==5.0 django-celery-beat` --> Success but with `django-celery-beat` downgraded to v2.1.0 and `django-timezone-field` downgraded from `v6.1.0` to `v4.2.3`

Please release a new version to PyPI.

hansegucker:
A release would be very great to bring forward our own release.

tuky:
Of course, we're all free to (temporarily) exclude this package from our requirements locking pipeline and install our deps something like this:

```sh
.../pipenv install --system --deploy
.../pip install python-crontab>=2.3.4 cron-descriptor>=1.2.32
.../pip install django-celery-beat==2.5.0 --no-dependencies
```

JulianTenschert:
> % `python3.12 -m venv .venv` % `source .venv/bin/activate` % `pip install --upgrade pip` % `pip install django==5.0 django-celery-beat==2.5.0` --> **_Failure_** % `pip install django==5.0 django-celery-beat` --> Success but with
> 
> * `django-celery-beat` downgraded `2.5.0` --> `v2.1.0` and
> * `django-timezone-field` downgraded from `v6.1.0`--> `v4.2.3`
> 
> Please release a new version to PyPI.

Thanks, this is a good workaround.

cclauss:
The tests pass on CPython and Django v5 but they fail on PyPy3.10 and Django v5.  Help needed.
 * https://github.com/celery/django-celery-beat/pull/705

Running on old versions of `django-celery-beat` and `django-timezone-field` is probably not a good idea.

ddoroshev:
[...]

PR #708, opened 2024-01-03
Fix Django 5.0 with PyPy 3.10

ddoroshev:
This PR fixes the problem of generating redundant migrations in PyPy 3.10/Django 5.0 environment.

In Django 5.0 they changed the way of normalizing choices, using [`match`](https://github.com/django/django/blob/45f778eded9dff59cfdd4dce8720daf87a08cfac/django/utils/choices.py#L76-L105).

CPython works fine with `case Iterable()`, so the choices always go through [this list comprehension](https://github.com/django/django/blob/45f778eded9dff59cfdd4dce8720daf87a08cfac/django/utils/choices.py#L109), being converted from tuples to lists every time.

PyPy 3.10-7.3.13 has a bug, it can't match `case Iterable()`, but works with `case Iterable` instead (which is wrong). So the test was failing, because the rendered schema contained `choices=[..., ...]`, while schema generated from the models containted `choices=(..., ...)`, which caused a redundant auto generated migration.

PyPy [fixed](https://foss.heptapod.net/pypy/pypy/-/commit/461b1615847a16570ec820b6562e188f80e8d73c) it in 7.3.14 though.

[Changelog](https://doc.pypy.org/en/latest/release-v7.3.14.html#id3):
> ### Python 3.10
>  <...>
>  - Pattern matching classes now use the full `isinstance` machinery, calling `__instancecheck__` too. ([#4036](https://foss.heptapod.net/pypy/pypy/-/issues/4036))

(The issue itself is unavailable, probably because they're moving to GitHub.)

So this small change of converting choices to lists fixes the problem.

Related issues and PRs:
- #698 
- #699 
- #705

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -27,21 +27,21 @@ MINUTES = 'minutes'
 SECONDS = 'seconds'
 MICROSECONDS = 'microseconds'
 
-PERIOD_CHOICES = (
+PERIOD_CHOICES = [
     (DAYS, _('Days')),
     (HOURS, _('Hours')),
     (MINUTES, _('Minutes')),
     (SECONDS, _('Seconds')),
     (MICROSECONDS, _('Microseconds')),
-)
+]
 
-SINGULAR_PERIODS = (
+SINGULAR_PERIODS = [
     (DAYS, _('Day')),
     (HOURS, _('Hour')),
     (MINUTES, _('Minute')),
     (SECONDS, _('Second')),
     (MICROSECONDS, _('Microsecond')),
-)
+]
 
 SOLAR_SCHEDULES = [
     ("dawn_astronomical", _("Astronomical dawn")),
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,6 +25,7 @@ Darjus Loktevic <darjus@amazon.com>
 David Fischer <david.fischer.ch@gmail.com>
 David Ziegler <david.ziegler@gmail.com>
 Diego Andres Sanabria Martin <diegueus9@gmail.com>
+Dima Doroshev <dima@doroshev.com>
 Dmitriy Krasilnikov <krasilnikov.d.o@gmail.com>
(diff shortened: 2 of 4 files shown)


django_celery_beat/models.py
```diff
@@ -27,21 +27,21 @@
...
 
-SINGULAR_PERIODS = (
+SINGULAR_PERIODS = [
     (DAYS, _('Day')),
     (HOURS, _('Hour')),
     (MINUTES, _('Minute')),
     (SECONDS, _('Second')),
     (MICROSECONDS, _('Microsecond')),
-)
+]
```
cclauss, line 44: Nice work!!

Can we please change this only on PyPy and in a way that allows it to be removed after PyPy v7.3.14 is released?

Tuples take less memory and signal to the reader that these values should normally not be modified at runtime.  Also, given comments in the first few episodes of the [core.py podcast](https://podcasters.spotify.com/pod/show/corepy), tuples should perform better in the multi-GIL world of Python 3.13+.

Something like...
```
import platform

if platform.python_implementation() == "PyPy":  # Workaround for PyPy < v7.3.14
    PERIOD_CHOICES = list(PERIOD_CHOICES)
    SINGULAR_PERIODS = list(SINGULAR_PERIODS)
```

cclauss (changes requested):
https://github.com/celery/django-celery-beat/pull/708/files#r1441435631

django_celery_beat/models.py
cclauss: Then why is there a bug at all?

ddoroshev: It's all about `test_models_match_migrations` that was failing in your PR https://github.com/celery/django-celery-beat/actions/runs/7267638477/job/19801878411?pr=705#step:6:2073

I think the PR description should cover the problem pretty well. If it's unclear, let me know and I'll fix it.

PR #717, opened 2024-01-19
fix long period task will never be triggered

daydaychen:
Here, the entry reload time is taken as the default value when last_run_at is None, 
and this value will be updated every time the schedule is updated/added. 

For a periodic task with a long period, such as three days, if there is always an update/add within three days, the periodic task will never be triggered.


daydaychen (commented):
The side effect is that adding a periodic task will trigger once first.

auvipy:
> The side effect is that adding a periodic task will trigger once first.

can you elaborate more on this please?

auvipy (changes requested):
we need proper unit test for the change to make sure there won't be any regression

daydaychen (commented):
> > The side effect is that adding a periodic task will trigger once first.

can you elaborate more on this please? 

When the `last_run_at` value is `None`, it defaults to the reload time of the entry.

For example, the first trigger time of a three-day interval scheduled task should be `last_run_at` +3 days, but each time the `schedule_changed` method return `True`, the entry will be re-instantiated and the `last_run_at` will be updated.

This can be a problem for the `schedule_changed` method frequently triggered by the management panel, because the `last_run_at` method will be frequently refreshed and may never be executed as planned.

Sorry, there were some side effects for the previous implementation. 

I made some changes to the initial value of `last_run_at`.

Now, if it's `None`, it will be fetched from `date_changed`.

In this way, the default value of `last_run_at` will not affected by the `schedule_changed`, the above problems can be solved.

daydaychen (commented):
> we need proper unit test for the change to make sure there won't be any regression 

Ok, I'll make up the unit test later.

auvipy (changes requested):
waiting for unit tests to make sure this will not create any regression

auvipy:
we need unit tests to make sure this won't create any regression/ unintended side effect

Akhilesh-Sirohi:
@daydaychen  any particular reason this PR isn't merged yet. I am facing the same [issue here](https://github.com/celery/django-celery-beat/issues/818) and thinking of implementing the same solution.

auvipy:
ugh! forgot to review it! going to review it right now!

auvipy (approved):

vkhudiakova:
Hi @auvipy! Do you know if there are any plans to release this change?

auvipy:
Yes in this month

PR #736, opened 2024-02-20
added sphinxcontrib-django to extensions

### LINKED ISSUE #735: Add `sphinxcontrib-django` to docs extensions
### Reasoning:

Right now the reference docs for models do nothing but list the fields while providing a generic description of a django field, and you can only really figure out what the fields do from the admin or by reading the source code. `sphinxcontrib-django` is an extension for autodoc which automatically documents django models based on their `help_text`. As you can see, before:

Screenshot 2024-02-20 at 5 53 32 PM

(ugly, useless) vs after:

Screenshot 2024-02-20 at 5 53 48 PM

(beautiful, informative). While I think the examples are good, there are some fields which aren't mentioned there that are nice to have described in the api reference. Obviously in order to provide the image reference I had to implement this, so I will open a pull request so others can check it out and browse the docs generated with the extension.

mfosterw:
Closes #735. To test this, first install the new dependency with `pip install requirements/docs.txt` and then build the docs normally.

### DIFF
--- a/docs/reference/django-celery-beat.models.rst
+++ b/docs/reference/django-celery-beat.models.rst
@@ -8,4 +8,3 @@
 
 .. automodule:: django_celery_beat.models
     :members:
-    :undoc-members:
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,4 +1,5 @@
 Django>=2.2,<5.0
+sphinxcontrib-django
 https://github.com/celery/sphinx_celery/archive/master.zip
 https://github.com/celery/kombu/zipball/main#egg=kombu
 https://github.com/celery/celery/zipball/main#egg=celery
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,6 +37,7 @@ globals().update(conf.build_config(
             None
         ),
     },
+    extensions=['sphinxcontrib_django']
 ))
 
 intersphinx_mapping = globals().get('intersphinx_mapping', {})


requirements/docs.txt
```diff
@@ -1,4 +1,5 @@
 Django>=2.2,<5.0
+sphinxcontrib-django
```
mfosterw, line 2: We can pin this to a specific version if necessary but I don't really see the need

docs/reference/django-celery-beat.models.rst
```diff
@@ -8,4 +8,3 @@
 
 .. automodule:: django_celery_beat.models
     :members:
-    :undoc-members:
```
mfosterw, line 11: This directive causes the fields documented by the extension to be listed twice

cclauss:
The only problem that I see is that this dependency [does not yet support Django v5](https://github.com/sphinx-doc/sphinxcontrib-django/blob/f4dfbbd382e6bb244d0f9ac7ee0b8930cff64c5e/pyproject.toml#L13-L15) and we want to put out a release that does support Dj5.  Could you please check to see if that will present problems?

mfosterw:
Setting the minimum version of Django to 5.0 did not cause any problems installing or running this for me, and since it's only a docs dependency I assume it shouldn't affect users directly.

cclauss (changes requested):
Blocked by sphinx-doc/sphinxcontrib-django#66

50-Course:
Thank you for this patch @mfosterw, while the former patch to this, does look informative enough, the latter shows great ease of use for developers. 

That said we can only push for https://github.com/sphinx-doc/sphinxcontrib-django/pull/66 to get merged.

Once again thank you @cclauss, @mfosterw.

Good job!

cclauss (approved):

Nusnus (approved):
Very nice, thank you!

PR #760, opened 2024-06-25
some fixes in fr translation

arthru:
image

There was a weird translation in french : "1 taĉhe a été fonctionnée avec succès"

In fact, there was a bad use of gettext for pluralization, so I used ngettext in place to ease the plural translation.

I also saw that there were some fuzzy translation in fr so I improved some translation.

I also refactored the calling method in order to keep the verb of the translatable string inside


django_celery_beat/locale/fr/LC_MESSAGES/django.po
```diff
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-12-22 19:03+0000\n"
+"POT-Creation-Date: 2024-06-25 11:24+0200\n"
 "PO-Revision-Date: 2022-10-14 23:48+0200\n"
```
cclauss, line 11: Should we be modifying the revision date instead of the creation date?

arthru: I just left what `django-admin makemessages -l fr` outputed but it wasn't a big change so I just rolled back that

thedrow (approved):

PR #764, opened 2024-07-03
Fix Tasks list in admin

zbig17:
The list of available tasks in admin (`TaskSelectWidget`) is based on currently loaded tasks. But celery `autodiscover_tasks` is lazy: "the actual auto-discovery won't happen until an application imports the default modules", ie. in worker.

This fix ensures that all tasks will be listed in `TaskSelectWidget`.


cclauss:
Can you please add so screen shots and/or tests that prove that this does something?

zbig17:
Sorry, `TaskSelectWidget._modules` does the same as I wanted to add.

PR #769, opened 2024-07-12
fix/clarify update_changed after TZ change

bbilly1:
I've stumbled over this on the docs website [here](https://django-celery-beat.readthedocs.io/en/latest/#important-warning-about-time-zones).

The code example there will raise `TypeError: PeriodicTasks.changed() missing 1 required positional argument: 'instance'`, as that expects the `PeriodicTask` instance as an argument, probably in a for loop as in the Readme [here](https://github.com/celery/django-celery-beat?tab=readme-ov-file#important-warning-about-time-zones).

I've set it to `update_changed`, I assume that will do the same as looping through the PeriodicTask instances and calling `changed` on each?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,8 +41,7 @@ Important Warning about Time Zones
 
         >>> from django_celery_beat.models import PeriodicTask, PeriodicTasks
         >>> PeriodicTask.objects.all().update(last_run_at=None)
-        >>> for task in PeriodicTask.objects.all():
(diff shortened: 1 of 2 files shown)


auvipy (approved):

bbilly1:
Very nice, thanks for merging. Looking forward to a docs rebuild, so future people aren't confused by this as well.

PR #778, opened 2024-07-30
Pre-commit: Add Python linter Ruff

cclauss:
https://docs.astral.sh/ruff

A more gradual approach to:
* #707

### DIFF
--- a/django_celery_beat/admin.py
+++ b/django_celery_beat/admin.py
@@ -24,8 +24,8 @@ class TaskSelectWidget(Select):
 
     def tasks_as_choices(self):
         _ = self._modules
-        tasks = list(sorted(name for name in self.celery_app.tasks
-                            if not name.startswith('celery.')))
+        tasks = sorted(name for name in self.celery_app.tasks
+                       if not name.startswith('celery.'))
         return (('', ''), ) + tuple(zip(tasks, tasks))
 
     @property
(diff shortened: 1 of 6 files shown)


django_celery_beat/admin.py
```diff
@@ -24,8 +24,8 @@ class TaskSelectWidget(Select):
 
     def tasks_as_choices(self):
         _ = self._modules
-        tasks = list(sorted(name for name in self.celery_app.tasks
-                            if not name.startswith('celery.')))
+        tasks = sorted(name for name in self.celery_app.tasks
```
auvipy, line 27: I'm not fully sure about this change

auvipy (changes requested):

django_celery_beat/admin.py
cclauss: It is unnecessary to use `list()` around `sorted()`, as the latter already returns a `list`.

Running the following on the commandline...

% [`ruff rule C413`](https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4)
# unnecessary-call-around-sorted (C413)

Derived from the **flake8-comprehensions** linter.

Fix is always available.

## What it does
Checks for unnecessary `list` or `reversed` calls around `sorted`
calls.

## Why is this bad?
It is unnecessary to use `list` around `sorted`, as the latter already
returns a list.

[ ... ]

PR #813, opened 2024-10-14
Add Python 3.13 to the testing

cclauss:
* https://www.python.org/downloads/release/python-3130/
* https://pythoninsider.blogspot.com/2024/10/python-3130-final-released.html
* https://devguide.python.org/versions/
* https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,18 +13,20 @@ jobs:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
         django-version: ["3.2", "4.2", "5.0", "5.1"]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy-3.10']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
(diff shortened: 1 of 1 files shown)


Nusnus (approved):
Thank you!

.github/workflows/test.yml
```diff
@@ -13,18 +13,20 @@ jobs:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
         django-version: ["3.2", "4.2", "5.0", "5.1"]
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', 'pypy-3.10']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
```
auvipy, line 16: I think support for python 3.8 is prematurely dropped as we are going to support python 3.8 with celery 5.5

auvipy, line 16: also inclusion of python 3.13 was not done across the board

cclauss: As the title says, this pr adds it to the tests, but does not declare full compatibility.

PR #835, opened 2025-01-02
refactor(db-scheduler): optimize all_as_schedule query

### LINKED ISSUE #675: Scheduler is slow when dealing with lots of tasks
### Summary:

When there are a few thousand tasks in the config, beat becomes completely unreliable and unstable. Reconfiguration, syncing takes a lot of time, and with frequent schedule changes, hours can pass by without sending any task to the queue.

* Celery Version: 5.3.4
* Celery-Beat Version: 2.5.0

1. Register 10k tasks with any config, (we only use CronSchedule)
2. Apply a change in the configuration in random intervals, 5s-300s, e.g. enabling or disabling tasks, or change the cron schedule.
3. Check the logs, you will see a lot of:
```
[2023-09-04 12:54:00,006: DEBUG/MainProcess] beat: Synchronizing schedule...

[2023-09-04 12:54:00,006: DEBUG/MainProcess] Writing entries...

[2023-09-04 12:55:49,129: DEBUG/MainProcess] beat: Synchronizing schedule...

[2023-09-04 12:55:49,130: DEBUG/MainProcess] Writing entries...
```
and beat is not applying any tasks.

### Detailed information

A few parts are clearly written without performance considerations, like:
`def all_as_schedule(...)`
could use the prefetch related api https://docs.djangoproject.com/en/4.2/ref/models/querysets/#prefetch-related 
`def sync(...)`
could use the bulk update api https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.query.QuerySet.bulk_update

Actually we tried to run a forked version with the above modifications, but it was just not enough, it is still very slow and unreliable.
Do you have any general ideas what should be the correct way to continue the investigation?

This "Writing entries..." part takes most of the time, and it is not *clearly* database related, I can see a lot of `SELECT "django_celery_beat_periodictasks"."ident", "django_celery_beat_periodictasks"."last_update" FROM "django_celery_beat_periodictasks" WHERE "django_celery_beat_periodictasks"."ident" = 1 LIMIT 21` which I assume is `def last_change(...)` , which is called in `def schedule_changed(...)` which is called from the `@property def schedule(...)` which is a bit harder to track.

- Are you intested in applying performance related optimisations?
- Should we run multiple instances of beat, maybe with slices of the whole schedule? Are there any caveats?
- Any other thoughts?

### LINKED ISSUE #786: DatabaseScheduler is bad for a large amount of tasks?
### Summary:

Hello, I am using celery beat to work with a large amount of future tasks in the database, It is starting to struggle when the number of tasks enabled is 100k+ (there are hundreds of thousands of disabled tasks)

The tasks do execute but after their scheduled time by a significant amount (few hours) which is not good for my use case, I tried scaling up the server it helped a little but the issue still exists and the number of tasks is going to increase over time.
Also increasing the number of workers doesn't seem to affect the issue very much.

Am I doing something wrong or is this the limit of the tool? any alternative solutions are very appreciated.

* Celery Version:  5.3.6
* Celery-Beat Version: 2.5.0

1.  Create thousands of tasks with different schedules.
2.  Wait for the tasks to execute on time
3.  Notice they are delayed from the scheduled time.

alirafiei75:
### Summary
This PR optimizes the `all_as_schedule` function to improve performance by filtering unnecessary tasks from the database query.

### Changes
- Exclude clocked tasks not due within 5 minutes.
- Exclude crontab tasks outside the relevant hour range.

fixes #675 
fixes #786

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -254,9 +254,13 @@ class DatabaseScheduler(Scheduler):
         debug('DatabaseScheduler: Fetching database schedule')
         s = {}
         next_five_minutes = now() + datetime.timedelta(minutes=5)
-        exclude_clock_tasks_query = Q(clocked__isnull=False, clocked__clocked_time__gt=next_five_minutes)
+        exclude_clock_tasks_query = Q(
+            clocked__isnull=False, clocked__clocked_time__gt=next_five_minutes
+        )
         exclude_hours = self.get_excluded_hours_for_crontab_tasks()
-        exclude_cron_tasks_query = Q(crontab__isnull=False, crontab__hour__in=exclude_hours)
+        exclude_cron_tasks_query = Q(
+            crontab__isnull=False, crontab__hour__in=exclude_hours
+        )
         for model in self.Model.objects.enabled().exclude(
             exclude_clock_tasks_query | exclude_cron_tasks_query
         ):
@@ -409,4 +413,7 @@ class DatabaseScheduler(Scheduler):
         }
 
         # Filter out 'should be considered' hours
-        return [hour for hour in allowed_crontab_hours if hour not in hours_to_remove]
+        return [
+            hour for hour in allowed_crontab_hours
+            if hour not in hours_to_remove
+        ]
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -410,6 +410,7 @@ class DatabaseScheduler(Scheduler):
             str(current_hour).zfill(2), str(current_hour),
             str(next_hour).zfill(2), str(next_hour),
             str(previous_hour).zfill(2), str(previous_hour),
+            str(4), "04",  # celery's default cleanup task
         }
 
         # Filter out 'should be considered' hours
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -11,14 +11,16 @@ from celery.utils.time import maybe_make_aware
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import close_old_connections, transaction
+from django.db.models import Q
 from django.db.utils import DatabaseError, InterfaceError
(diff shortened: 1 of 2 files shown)


alirafiei75:
Screenshot from 2025-01-04 12-04-45

This is cpu usage of our celery-beat server. We depend highly on cronjobs and clockjobs and as you see, optimizing the scheduler query, led to reduction in cpu usage considerably (see 11:53 in the picture)

auvipy (commented):
The changes look good from the first review.  I will thoroughly Review it again in a couple of days before merging

django_celery_beat/schedulers.py
```diff
@@ -251,7 +253,17 @@ def setup_schedule(self):
     def all_as_schedule(self):
```
auvipy, line 253: can we add/modify additional test for this change?

django_celery_beat/schedulers.py
```diff
@@ -377,3 +389,32 @@ def schedule(self):
                     repr(entry) for entry in self._schedule.values()),
                 )
         return self._schedule
+
+    @staticmethod
```
auvipy, line 393: and for this method?

auvipy (changes requested):

django_celery_beat/schedulers.py
alirafiei75: I have added tests for this method too

auvipy (approved):

auvipy:
thanks for your work. feel free to contribute back if you find any better improvement or regression.

auvipy:
hey, we got report about this PR https://github.com/celery/django-celery-beat/issues/875#issuecomment-2821533541 can you please check and if possible come with a possible fix? otherwise we have to revert this

auvipy:
I can wait for your pr

alirafiei75:
I have made the PR @auvipy . please review it whenever it is convenient #879

PR #838, opened 2025-01-14
Update README.rst

angryfoxx:


cclauss:
My sense is that the current text is correct and this pull request is not needed.

PR #844, opened 2025-02-09
Fix the time calculation problem caused by start_time

### LINKED ISSUE #843: Celery Task with start_time Set in the Future Causes Blocking of Other Tasks
## Description:

I encountered an issue when setting a specific `start_time` for a scheduled task in `django-celery-beat`. If the `start_time` is set to a time in the future, Celery will treat this time as part of the `event_t` tuple, specifically the `event_t.time` field. Celery continues to check whether the task has reached its execution condition.

However, if the `start_time` is greater than the current time and the time for the next `crontab` run has not yet been reached, the task will continue to be added to the heap. As a result, the `tick` function continuously processes this task in the loop, causing it to block the execution of other tasks.

This kind of exception is unlikely to occur with Celery itself, but when used in conjunction with django-celery-beat, it is easy to reproduce the issue due to the user-defined start_time setting.

## Code with Explanation:

```python
    def tick(self, event_t=event_t, min=min, heappop=heapq.heappop,
             heappush=heapq.heappush):
        """Run a tick - one iteration of the scheduler.

        Executes one due task per call.

        Returns:
            float: preferred delay in seconds for next call.
        """
        adjust = self.adjust
        max_interval = self.max_interval

        # If the heap is empty or the scheduled tasks have changed, repopulate the heap.
        if (self._heap is None or
                not self.schedules_equal(self.old_schedulers, self.schedule)):
            self.old_schedulers = copy.copy(self.schedule)
            self.populate_heap()

        H = self._heap

        # If the heap is empty, return the maximum interval.
        if not H:
            return max_interval

        event = H[0]
        entry = event[2]
        is_due, next_time_to_run = self.is_due(entry)

        # If the task is due, process the task.
        if is_due:
            verify = heappop(H)
            if verify is event:
                next_entry = self.reserve(entry)
                self.apply_entry(entry, producer=self.producer)

                # Update the task's next scheduled time and push it back into the heap
                heappush(H, event_t(self._when(next_entry, next_time_to_run),
                                    event[1], next_entry))

                return 0
            else:
                # If the task has been modified, push it back into the heap and return the next shortest delay
                heappush(H, verify)
                return min(verify[0], max_interval)

        # **Explanation of the Core Issue**:
        # When `is_due` is False, the task should not be executed immediately. 
        # However, if the task's `event.time` is the smallest in the heap, it will remain at the top of the heap.
        # This is because `event.time` is smaller than the times of other tasks.
        # Since the subsequent processing doesn't adjust this task's `event.time`, 
        # the task will keep being processed and block other tasks from being executed.
        
        adjusted_next_time_to_run = adjust(next_time_to_run)

        # If `next_time_to_run` is a valid numeric value, return the next execution time.
        return min(adjusted_next_time_to_run if is_numeric_value(adjusted_next_time_to_run) else max_interval,
                   max_interval)
 ``` 

## Steps to Reproduce:

1. Set a `start_time` in the future for a periodic task.
2. Set up the task with a `crontab` schedule.
3. Set another periodic task with a 10-second interval.
4. Observe that when the current time is earlier than the `start_time` and the next `crontab` execution time has not yet arrived, the task continues to be added to the heap.
5. Notice that when the `event.time` generated by the `crontab` task is the smallest, it will remain at the top of the heap, blocking the execution of other tasks (such as the 10-second interval task).

## Expected Behavior:

The task should not appear at the top of the heap repeatedly before its execution time, regardless of the schedule. This would prevent unnecessary checks and blocking of other tasks that should run in parallel.

## Possible Solution:

The issue can be addressed by modifying the `is_due` method in `django_celery_beat/schedulers.py`. The method should calculate the next valid scheduled time based on the `start_time` and task schedule `crontab`. Tasks should only be checked for execution when this time is reached.

[link](https://github.com/celery/django-celery-beat/pull/844/files)

Since I don't know how to set up a unit testing environment for django-celery-beat, I used manual simulation testing. However, I'm unsure if my code modification is correct and whether it can contribute to the community.

Alternatively, the `tick` method in Celery could be updated to handle tasks with a future `event_t.time`, preventing them from blocking other tasks until they are ready to execute.

Either solution would prevent tasks with a future `event_t.time` from being added to the heap and continuously processed, blocking other tasks.

## Environment:

celery==5.4.0
Django==5.1.6
django-celery-beat==2.7.0
python==3.12

### PR DESCRIPTION
**Pull request by JinRiYao2001**
### Description  
This PR fixes a critical scheduling issue in `django-celery-beat` when a **crontab-type task** is configured with a `start_time` set to a future time.  

#### Problem Analysis  
- When initializing the scheduler's heap (`self._heap`), tasks are sorted by their `event_t.time` field.  
- For crontab tasks with `start_time > now`, the initial `event_t.time` was incorrectly set to `start_time` instead of the **next valid execution time** based on the crontab schedule.  
- This caused the task to persistently occupy the heap's top position (since `start_time` was the smallest future timestamp), blocking other tasks from being executed until the `start_time` was reached.  

#### Solution  
- **Modified the `event_t.time` generation logic** during heap initialization for crontab tasks with future `start_time` values.  
- Instead of directly using `start_time`, the `event_t.time` is now calculated as the **next valid execution time** derived from the crontab schedule **relative to `start_time`**.  
  - Example: If a task is set to run every hour (`0 * * * *`) with `start_time=2024-01-01 12:00:00` (future time), the initial `event_t.time` will be `2024-01-01 12:00:00` (first valid run) instead of the current time.  
- Ensures the heap prioritizes tasks based on their **actual next execution time** rather than `start_time`, preventing heap blockage.  

#### Changes
- Enhanced the `is_due` method in `schedulers.py` to:
  - Add special handling for tasks with non-null `start_time`
  
#### Expected Behavior  
- Tasks with future `start_time` will no longer block the scheduler heap.  
- The heap now correctly prioritizes tasks based on their dynamically calculated next execution time, allowing other interval/crontab tasks to execute as expected.  

fixes https://github.com/celery/django-celery-beat/issues/843

### DIFF
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -390,8 +390,8 @@ class CrontabSchedule(models.Model):
         except MultipleObjectsReturned:
             return cls.objects.filter(**spec).first()
 
-    def due_start_time(self, start_time, tz):
-        start_time = start_time.astimezone(tz)
+    def due_start_time(self, initial_start_time, tz):
+        start_time = initial_start_time.astimezone(tz)
         start, ends_in, now = self.schedule.remaining_delta(start_time)
         return start + ends_in
 
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -391,8 +391,7 @@ class CrontabSchedule(models.Model):
             return cls.objects.filter(**spec).first()
 
     def due_start_time(self, start_time, tz):
-        if str(start_time.tzinfo) != str(tz):
-            start_time = start_time.astimezone(tz)
+        start_time = start_time.astimezone(tz)
         start, ends_in, now = self.schedule.remaining_delta(start_time)
         return start + ends_in
 
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -663,3 +663,11 @@ class PeriodicTask(models.Model):
     @property
     def schedule(self):
         return self.scheduler.schedule
+
+    @property
+    def due_start_time(self):
+        if self.crontab:
+            _, delay = self.scheduler.schedule.is_due(self.start_time)
+            return self.start_time + timedelta(seconds=delay)
+        else:
+            return self.start_time
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -392,9 +392,13 @@ class CrontabSchedule(models.Model):
 
     def due_start_time(self, start_time):
         start, ends_in, now = self.schedule.remaining_delta(start_time)
-        if (str(start.tzinfo) == str(now.tzinfo) and
-            now.utcoffset() != start.utcoffset()):
+
+        same_tz = str(start.tzinfo) == str(now.tzinfo)
+        different_offset = now.utcoffset() != start.utcoffset()
+
+        if same_tz and different_offset:
             start = start.replace(tzinfo=now.tzinfo)
+
         return start + ends_in
 
 
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -390,6 +390,13 @@ class CrontabSchedule(models.Model):
         except MultipleObjectsReturned:
             return cls.objects.filter(**spec).first()
 
+    def due_start_time(self, start_time):
+        start, ends_in, now = self.schedule.remaining_delta(start_time)
+        if str(start.tzinfo) == str(
+            now.tzinfo) and now.utcoffset() != start.utcoffset():
+            start = start.replace(tzinfo=now.tzinfo)
+        return start + ends_in
+
 
 class PeriodicTasks(models.Model):
     """Helper table for tracking updates to periodic tasks.
@@ -667,7 +674,6 @@ class PeriodicTask(models.Model):
     @property
     def due_start_time(self):
         if self.crontab:
-            _, delay = self.scheduler.schedule.is_due(self.start_time)
-            return self.start_time + timedelta(seconds=delay)
+            return self.crontab.due_start_time(self.start_time)
         else:
             return self.start_time
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -392,8 +392,8 @@ class CrontabSchedule(models.Model):
 
     def due_start_time(self, start_time):
         start, ends_in, now = self.schedule.remaining_delta(start_time)
-        if str(start.tzinfo) == str(
-            now.tzinfo) and now.utcoffset() != start.utcoffset():
+        if (str(start.tzinfo) == str(now.tzinfo) and
+            now.utcoffset() != start.utcoffset()):
             start = start.replace(tzinfo=now.tzinfo)
         return start + ends_in
 
--- a/django_celery_beat/models.py
+++ b/django_celery_beat/models.py
@@ -390,15 +390,10 @@ class CrontabSchedule(models.Model):
         except MultipleObjectsReturned:
             return cls.objects.filter(**spec).first()
 
-    def due_start_time(self, start_time):
+    def due_start_time(self, start_time, tz):
+        if str(start_time.tzinfo) != str(tz):
+            start_time = start_time.astimezone(tz)
         start, ends_in, now = self.schedule.remaining_delta(start_time)
-
-        same_tz = str(start.tzinfo) == str(now.tzinfo)
-        different_offset = now.utcoffset() != start.utcoffset()
-
-        if same_tz and different_offset:
-            start = start.replace(tzinfo=now.tzinfo)
-
         return start + ends_in
 
 
@@ -675,9 +670,8 @@ class PeriodicTask(models.Model):
     def schedule(self):
         return self.scheduler.schedule
 
-    @property
-    def due_start_time(self):
+    def due_start_time(self, tz):
         if self.crontab:
-            return self.crontab.due_start_time(self.start_time)
+            return self.crontab.due_start_time(self.start_time, tz)
         else:
             return self.start_time
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -117,9 +117,8 @@ class ModelEntry(ScheduleEntry):
                 # The datetime is before the start date - don't run.
                 # send a delay to retry on start_time
 
-                next_time_run = self.schedule.remaining_estimate(
-                    self.model.start_time)
-                delay = math.ceil(next_time_run.total_seconds())
+                delay = math.ceil(
+                    (self.model.due_start_time - now).total_seconds())
 
                 return schedules.schedstate(False, delay)
 
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -116,7 +116,7 @@ class ModelEntry(ScheduleEntry):
             if now < self.model.start_time:
                 # The datetime is before the start date - don't run.
                 # send a delay to retry on start_time
-                now_tz = now.tzinfo
+                current_tz = now.tzinfo
                 delay = math.ceil(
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by JinRiYao2001** [comment]
## Changes:
- Modified the `delay` calculation logic in `is_due()`
- Now, `delay` is calculated based on `start_time`, `now`, and `schedule.remaining_estimate()`
- This ensures a more accurate computation of the next execution time

## Why is this change needed?
- Previously, `delay` was calculated as `start_time - now`, which is not always accurate
- The new logic ensures that the delay aligns with the scheduler's expected execution time

## Impact:
- The unit test `test_task_with_start_time` may fail due to the updated `delay` calculation
- The test still assumes `delay = start_time - now`, which does not match the new logic
- It is necessary to update the test to align with the `schedule.remaining_estimate()` calculation

## Suggested Solution:
- I propose updating `test_task_with_start_time` to validate the new `delay` calculation logic

## Question  
- Should I submit an updated unit test along with this change?

**Comment by Nusnus** [comment]
_Changed to draft until the CI passes_

**Comment by JinRiYao2001** [comment]
I'll take the time to fix this code

**Comment by JinRiYao2001** [comment]
I discovered a bug in my code while trying to write unit tests. I will fix it later and conduct the corresponding tests.

**Comment by JinRiYao2001** [comment]
My code still has some time zone issues, I will modify it tomorrow

**Comment by JinRiYao2001** [comment]
I think I solved the time zone issue.

**Review by auvipy — CHANGES_REQUESTED** [review]

File: requirements/test.txt [hunk 1]
```diff
@@ -2,3 +2,4 @@ pytest-django>=4.5.2,<5.0
 pytest>=6.2.5,<9.0
 pytest-timeout
 ephem
+pytz
```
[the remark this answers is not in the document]
JinRiYao2001: Thank you for your feedback! I've replaced the usage of pytz with zoneinfo as suggested. I appreciate your recommendation and have updated the code accordingly to make use of the standard library. Please let me know if there’s anything else you'd like me to adjust.

Thanks again!

django_celery_beat/models.py
```diff
@@ -390,6 +390,12 @@ def from_schedule(cls, schedule):
         except MultipleObjectsReturned:
             return cls.objects.filter(**spec).first()
 
+    def due_start_time(self, start_time, tz):
+        if str(start_time.tzinfo) != str(tz):
```
cclauss, line 394: Is string comparison the right approach?  `ZoneInfo("America/New_York") != ZoneInfo("America/Toronto")` but they are the same timezone.  This could be replicated in other values in `zoneinfo.available_timezones()`.  Should we just calculate a time_delta instead of comparing strings?  I always find that working with datetime is difficult so if @auvipy is happy with the current approach then so am I.

JinRiYao2001: Thank you for catching this nuance! You're absolutely right that string comparison isn't reliable for timezone equivalence. While I took inspiration from [Celery's implementation pattern](https://github.com/celery/celery/blob/main/celery/utils/time.py#L224), I realize direct object comparison (like `start_time.tzinfo == tz`) would still fail in cases of equivalent-but-different zone instances. I also think that performing excessive calculations and comparisons to check if the time zones are the same is less efficient than directly performing the conversion.

After reconsidering, the simplest approach is indeed to eliminate comparisons altogether. I've refactored the code to simply perform unconditional timezone conversion:

```python
def due_start_time(self, start_time, tz):
    start_time = start_time.astimezone(tz)
    start, ends_in, now = self.schedule.remaining_delta(start_time)
    return start + ends_in
```
This avoids ambiguous checks while guaranteeing correct timezone alignment through Python's built-in astimezone() handling.

Please let me know if further adjustments would be helpful! 😀

auvipy:
I will re review it tomorrow

t/unit/test_schedulers.py
```diff
@@ -777,6 +778,122 @@ def test_starttime_trigger(self, monkeypatch):
         assert s._heap[0]
         assert s._heap[0][2].name == m1.name
 
+    def test_crontab_with_start_time_between_now_and_crontab(self, app):
+        now = app.now()
+
```
auvipy, line 783:
```suggestion
```

auvipy (commented):
would you mind updating the PR description which fits the latest changes and finding from your work, please? also do you thin this should work on python 3.8?

JinRiYao2001:
@auvipy 
I've updated the PR description accordingly. Regarding Python 3.8 compatibility, I've only used the ZoneInfo library which should maintain compatibility. I'll verify this by running the test cases in a Python 3.8 environment locally to be certain.

JinRiYao2001:
@auvipy 
I've updated the PR description accordingly and reviewed the compatibility concerns and confirmed that my code currently only uses ZoneInfo in test cases. I'll address this by implementing a fix to ensure full compatibility with Python 3.8.

auvipy:
thanks! that's neat!

django_celery_beat/schedulers.py
```diff
@@ -116,9 +116,11 @@ def is_due(self):
             if now < self.model.start_time:
                 # The datetime is before the start date - don't run.
                 # send a delay to retry on start_time
+                now_tz = now.tzinfo
```
auvipy, line 119:
```suggestion
                current_tz = now.tzinfo
```

django_celery_beat/schedulers.py
```diff
@@ -116,9 +116,11 @@ def is_due(self):
...
                 delay = math.ceil(
-                    (self.model.start_time - now).total_seconds()
+                    (self.model.due_start_time(now_tz) - now).total_seconds()
```
auvipy, line 121:
```suggestion
                    (self.model.due_start_time(current_tz) - now).total_seconds()
```

auvipy (changes requested):

django_celery_beat/models.py
```diff
@@ -390,6 +390,11 @@ def from_schedule(cls, schedule):
         except MultipleObjectsReturned:
             return cls.objects.filter(**spec).first()
 
+    def due_start_time(self, start_time, tz):
```
auvipy, line 393: can you please Consider renaming the parameter 'start_time' to 'initial_start_time' to more clearly indicate its role in calculating the next valid execution time?

auvipy (changes requested):

auvipy:
django_celery_beat/schedulers.py:121:80: E501 line too long (81 > 79 characters)

auvipy:
related https://github.com/celery/django-celery-beat/pull/867

django_celery_beat/models.py
auvipy: if you can address this by tomorrow, it will be great. we are very close to merge it

JinRiYao2001: @auvipy 
Can we consider the Python 3.8 compatibility issues resolved after successfully installing both requirements/test.txt and requirements/default.txt, and having all test cases pass?

auvipy:
Thanks for incorporating new pytest timeout features

django_celery_beat/models.py
auvipy: Yes

auvipy (approved):

auvipy:
hey, we are having regression report related to this pr. can you please check?

JinRiYao2001:
@auvipy 
Thank you for the feedback. The issue may be related to my PR, but due to a heavy workload, I’m unable to investigate thoroughly at the moment. I’ll look into it as soon as my schedule allows.

auvipy:
Thanks

auvipy:
it seems that it would be better to revert part of this

JinRiYao2001:
@auvipy Thanks, looking into it. Will post my findings on #1045.

PR #846, opened 2025-02-13
Add Romanian translation

bylly1:


auvipy (changes requested):
do we have additional romanian speaker who can review this?

PR #849, opened 2025-02-25
remove the last trace of "ephem"

a-detiste:
Hi,

I see usage was removed before


auvipy (changes requested):
aren't we using this?

PR #858, opened 2025-03-21
fix(tests): correct assertion in excluded hours test

kyonlee:
The `test_get_excluded_hours_for_crontab_tasks` method had assertions for excluded hours that could fail depending on the test execution time. We can only guarantee what hours are not returned by `test_get_excluded_hours_for_crontab_tasks` and have amended the test to do so.


kyonlee:
When the test is executed on the 2nd or 6th hour, the assertions checking `now_hour + 2` and `now_hour - 2` are in `excluded_hours` will fail respectively, because "4" and "04" are always removed from `excluded_hours`.

I have removed these assertions and have added one to verify that "4" and "04" is not in `excluded_hours`.

alirafiei75 (approved):

PR #863, opened 2025-03-27
Set all the python/django version numbers correctly

jrobichaud:
- [x] remove reference to django 3.2, 4.1, 4.2 (EOL)
- [x] remove python 3.8 (EOL)
- [x] replace pypy-3.10 with pypy-3.11
- [x] add django 5.2rc1 to tox
- [x] make sure tools runs with 3.9 
- [x] tell readthedocs to use latest python and ubuntu lts

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,7 +44,7 @@ jobs:
         python -m pip install tox tox-gh-actions
     - name: Test with tox
       run: |
-        if [ "${{ matrix.python-version }}" != "pypy-3.10" ]; then
+        if [ "${{ matrix.python-version }}" != "pypy-3.11" ]; then
           tox -- --cov=django_celery_beat --cov-report=xml --no-cov-on-fail --cov-report term
         else
           tox
@@ -52,7 +52,7 @@ jobs:
       env:
         DJANGO: ${{ matrix.django-version }}
     - name: Upload coverage reports to Codecov
-      if: ${{ matrix.python-version != 'pypy-3.10' }}
+      if: ${{ matrix.python-version != 'pypy-3.11' }}
       uses: codecov/codecov-action@v5
       with:
         fail_ci_if_error: true # optional (default = false)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
         django-version: ["4.2", "5.0", "5.1", "5.2rc1"]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.11']
         exclude:
           - django-version: "4.2"
             python-version: "3.13"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,15 +12,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
-        django-version: ["3.2", "4.2", "5.0", "5.1", "5.2rc1"]
+        django-version: ["4.2", "5.0", "5.1", "5.2rc1"]
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
         exclude:
-          - django-version: "3.2"
-            python-version: "3.11"
-          - django-version: "3.2"
-            python-version: "3.12"
-          - django-version: "3.2"
-            python-version: "3.13"
           - django-version: "4.2"
             python-version: "3.13"
           - django-version: "5.0"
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -12,15 +12,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
-        django-version: ["3.2", "4.2", "5.0", "5.1", "5.2rc1"]
+        django-version: ["4.2", "5.0", "5.1", "5.2rc1"]
```
cclauss, line 15: @auvipy Are you OK with dropping Django v3.2 support?

Would this fail if we had "5.2" instead of "5.2rc1"?

.readthedocs.yaml
```diff
@@ -2,7 +2,7 @@ version: 2
 build:
   os: ubuntu-20.04
   tools:
-    python: "3.8"
+    python: "3.12"
```
cclauss, line 5: Let's just use the current Python so we can spot errors with the current Python as they arise.
```suggestion
    python: "3.x"
```

.github/workflows/test.yml
```diff
@@ -12,15 +12,9 @@ jobs:
...
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
```
cclauss, line 16:
```suggestion
        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.11']
```

tox.ini
```diff
@@ -39,11 +38,11 @@ deps=
 
     cov: -r{toxinidir}/requirements/test-django.txt
 
-	django32: Django ~= 3.2
 	django41: Django ~= 4.1
 	django42: Django ~= 4.2
 	django50: Django ~= 5.0
-	django51: Django ~= 5.1rc1
+	django51: Django ~= 5.1
+	django52: Django ~= 5.2rc1
```
cclauss, line 45: Would this fail if we had "5.2" instead of "5.2rc1"?

tox.ini
```diff
@@ -2,29 +2,28 @@
...
     pypy-3.10: pypy3
 
 [gh-actions:env]
 DJANGO =
-    3.2: django32
-    4.1: django41
     4.2: django42
     5.0: django50
     5.1: django51
+    5.2rc1: django52
```
cclauss, line 17: Would this fail if we had "5.2" instead of "5.2rc1"?

cclauss (changes requested):

.github/workflows/test.yml
jrobichaud: For the moment yes.

tox.ini
jrobichaud: Yes.

tox.ini
jrobichaud: Yes

jrobichaud:
@cclauss when django 5.2 will be officially out we will be able to change 5.2rc1 to 5.2.

Pre-releases have to be exact matches.

.readthedocs.yaml
```diff
@@ -2,7 +2,7 @@ version: 2
 build:
   os: ubuntu-20.04
```
cclauss, line 3: GitHub Actions are just about to drop u20.
```suggestion
  os: ubuntu-latest
```

jrobichaud: i changed it to `ubuntu-lts-latest` like in their doc: https://docs.readthedocs.com/platform/stable/config-file/v2.html#build-os

.github/workflows/test.yml
```diff
@@ -12,15 +12,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # https://docs.djangoproject.com/en/stable/faq/install/#what-python-version-can-i-use-with-django
-        django-version: ["3.2", "4.2", "5.0", "5.1", "5.2rc1"]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.10']
+        django-version: ["4.2", "5.0", "5.1", "5.2rc1"]
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13', 'pypy-3.11']
```
cclauss, line 16: This fails!  Perhaps leave it for a separate pull request.

jrobichaud: Values have to be changed in both `tox.ini` and in `test.yml`

jrobichaud:
In a few days we will need to make a new PR replacing `5.2rc1` with `5.2`.

jrobichaud:
@cclauss, I let you resolve the conversations if you are satisfied with the fixes. When this one is merge, I have another one I will submit to be merged when django 5.2 will be out officially.

cclauss (approved):

auvipy (changes requested):
Im going to hold this

jrobichaud:
@auvipy , on hold like waiting for django 5.2 to release? 

If so, I will just replace the `5.2rc1` to `5.2` already.

auvipy (changes requested):
addressed by https://github.com/celery/django-celery-beat/pull/865

auvipy:
> @auvipy , on hold like waiting for django 5.2 to release?
> 
> If so, I will just replace the `5.2rc1` to `5.2` already.

we tend to keep bit older versions supported to make the transition easier for legacy projects. thanks for you efforts though. the CI was not fully synced so I have to take care of them.

Issue #874, opened 2025-04-18
ValueError: localtime() cannot be applied to a naive datetime


xiaobao0505:
There is an error when starting the project：
django-celery-beat: 2.8.0

1.celery -A app beat -l info
2.
![Image](https://github.com/user-attachments/assets/68500da7-0c92-4a98-a1bf-caa5d8fb186e)

xiaobao0505:
It should be a bug triggered when USE_TZ is set to False.

cristianfavaro:
Having the same issue. 
I've been using it as "USE_TZ = False" in Django settings and was ok. But now it is not working. Crazy.

auvipy:
can you guys help to track the commit causing this issue? any test to reproduce it?

djarv:
[Commit fc64741, django_celery_beat/schedulers.py, line 404](https://github.com/celery/django-celery-beat/commit/fc64741a720b4bbd09308df1553a47a3f0d979ab#diff-30f45e7eed4aafe79dbc3b004881b237faf4393938b769fff25325364384bcedR404) in the new` get_excluded_hours_for_crontab_tasks()` function.

`timezone.localtime` expects a tz aware timestamp, and `.utils.now()` will return naive timestamp if `USE_TZ` is `False`.
I have a local fork in dev replacing `now()` with `datetime.datetime.now(datetime.timezone.utc)`, which as I understand `timezone.localtime` can use to convert to whatever local time it is trying to do - not hitting TZ unaware database with this date. I'm not sure if there is a more official django/celerybeat way of doing what I've changed but it seems to be running again. This needs testing properly though and my server is UTC though so I'm not in a position to test other timezones.

Recreate: django setting `USE_TZ` is `False` (when running a TZ unaware db).  
`DJANGO_CELERY_BEAT_TZ_AWARE` setting also exists but won't cause this bug, though I assume things will fall apart if `USE_TZ` and `DJANGO_CELERY_BEAT_TZ_AWARE` aren't set the same?

auvipy:
Would you mind a draft pr with possible fix?

djarv:
draft PR [#877](https://github.com/celery/django-celery-beat/pull/877) created with the change I mentioned.

LT-lt12:
> It should be a bug triggered when USE_TZ is set to False.

确实存在此问题，你尝试解决了吗？我试图尝试

![Image](https://github.com/user-attachments/assets/53ba7958-66e0-45a6-9923-2a7398847478)

[...]

PR #877, opened 2025-04-21
Re. Issue #874: guarantee tz aware current datetime for timezone.localtime

djarv:
re. Issue #874, draft PR as per request.

`timezone.localtime` requires a timezone aware datetime, and `now` will return a naive datetime if setting `USE_TZ` is `False`, resulting in failed startup `ValueError: localtime() cannot be applied to a naive datetime`.

`USE_TZ` and `DJANGO_CELERY_BEAT_TZ_AWARE` settings should not be relevant to the output of method `DatabaseScheduler.get_excluded_hours_for_crontab_tasks`, which is a list of naive hour strings.

Maybe there is a more conforming way of accomplishing this than my suggestion, but cannot use `maybe_make_aware(now())` because `now()` returns naive local time, and `maybe_make_aware` requires naive utc time (e.g. from depricated `datetime.utcnow()`).

While I've had this running on my UTC timezone dev server for a few days, this is not tested, particularly for other timezones, and I do not claim to fully understand how the timezone/naive timestamp management works through these modules.

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -4,6 +4,11 @@ import logging
 import math
 from multiprocessing.util import Finalize
 
+try:
+    from zoneinfo import ZoneInfo  # Python 3.9+
+except ImportError:
+    from backports.zoneinfo import ZoneInfo  # Python 3.8
+
 from celery import current_app, schedules
 from celery.beat import ScheduleEntry, Scheduler
 from celery.utils.log import get_logger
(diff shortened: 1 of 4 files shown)


django_celery_beat/schedulers.py
```diff
@@ -403,7 +403,10 @@ def get_excluded_hours_for_crontab_tasks():
         ]
 
         # Get current, next, and previous hours
-        current_time = timezone.localtime(now())
+        current_time = timezone.localtime(
+            # timezone.localtime() expects a timezone aware datetime.
+            datetime.datetime.now(datetime.timezone.utc)
```
auvipy, line 408: would you mind investigate more places where this type of changes are needed?

auvipy (changes requested):

djarv:
Code refactor in #879 has removed the affected code that this PR was looking to address. Re-based my fork which closed this draft PR.

Issue #885, opened 2025-05-06
New behavior exclude specific crontab hours - It uses django's timezone rather than server


iwalucas:
There's an issue in django-celery-beat's DatabaseScheduler class where timezone calculations lead to incorrect task scheduling when the Celery timezone differs from the server timezone.

In our project, we have:

Server running in UTC
CELERY_TIMEZONE = "EST" in settings
Tasks scheduled using crontab hour values

**Problem**
The `_get_timezone_offset` method in `schedulers.py` gets the wrong server timezone information. It uses `timezone.get_current_timezone()` which returns a timezone-aware datetime, but then incorrectly interprets the timezone from it.

This leads to the scheduler making incorrect decisions about which tasks to run because the timezone offset calculation is wrong. Tasks scheduled in different timezones don't run at the expected times.

**Expected Behavior**
The scheduler should correctly calculate the timezone offset between the server timezone and each task's timezone, accounting for the Celery timezone setting.

**Steps to Reproduce**
Set CELERY_TIMEZONE = "EST" in Django settings
Run the server on a machine with UTC timezone
Schedule a task with a crontab for a specific hour
Observe that the task wont run at the defined time

**Suggested Fix**
The `_get_timezone_offset` method should account for the server timezone. This is how I fixed:
```
        server_time = aware_now()
        server_tz = ZoneInfo(str(server_time.tzinfo))
```

**Environment**
django-celery-beat version: 2.8.0
Django version: 5.2
Python version: 3.11

auvipy:
we have a new fix for this https://github.com/celery/django-celery-beat/pull/879, this should be released soon

iwalucas:
@auvipy that code doesnt fix this, thats why I raised this new issue.

auvipy:
Oh wow. Can you please come with a PR?  Thanks

iwalucas:
@auvipy yes, let me do a PR with my fix (not most elegant, but works)

iwalucas:
@auvipy Here is the PR [#886](https://github.com/celery/django-celery-beat/pull/886/files)

becky-gilbert:
I'm having the exact same problem reported here - thanks @iwalucas for the issue and fix. 
I pinpointed this to the same lines in `_get_timezone_offset`. And in case it's helpful, I came up with this solution: `server_time = timezone.localtime()`.

auvipy:
we got a pr merged to fix this issue. you can check that out too

auvipy:
[...]

PR #886, opened 2025-05-07
Consider server timezone on _get_timezone_offset instead of django's settings

iwalucas:
DatabaseScheduler class consider's django timezone calculations to offset hour, but beat uses server's so that creates a situation where some tasks are excluded from running hours

Report: [#885 ](https://github.com/celery/django-celery-beat/issues/885)

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -364,7 +364,9 @@ class DatabaseScheduler(Scheduler):
             int: The hour offset
         """
         # Get server timezone
-        server_tz = timezone.get_current_timezone()
+        # this gets django settings and not server's 
+        # server_tz = timezone.get_current_timezone()
+        server_tz = ZoneInfo(str(server_time.tzinfo))
 
         if isinstance(timezone_name, ZoneInfo):
             timezone_name = timezone_name.key
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -19,7 +19,6 @@ from django.db import close_old_connections, transaction
 from django.db.models import Case, F, IntegerField, Q, When
 from django.db.models.functions import Cast
 from django.db.utils import DatabaseError, InterfaceError
-from django.utils import timezone
 from kombu.utils.encoding import safe_repr, safe_str
 from kombu.utils.json import dumps, loads
 
@@ -364,8 +363,6 @@ class DatabaseScheduler(Scheduler):
             int: The hour offset
         """
         # Get server timezone
-        # this gets django settings and not server's 
-        # server_tz = timezone.get_current_timezone()
         server_time = aware_now()
         server_tz = ZoneInfo(str(server_time.tzinfo))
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
I was thinking if you can add some additional tests to verify the changes, and check if it creates any regresion

django_celery_beat/schedulers.py
```diff
@@ -364,7 +363,9 @@ def _get_timezone_offset(self, timezone_name):
             int: The hour offset
         """
         # Get server timezone
-        server_tz = timezone.get_current_timezone()
+        server_time = aware_now()
+        # Use server_time.tzinfo directly if it is already a ZoneInfo instance
+        server_tz = server_time.tzinfo if isinstance(server_time.tzinfo, ZoneInfo) else ZoneInfo(str(server_time.tzinfo))
```
jennifer-richards, line 368: The `tzinfo` class does not define `__str__`. It would be better to use `server_time.tzinfo.tzname()` which is guaranteed to exist.

There's no guarantee `ZoneInfo` will know what to do with the `tzname()`, though, so this won't work with obscure timezone implementations.

jennifer-richards (changes requested):
Suggest a minor rephrasing in the code.

auvipy:
@alirafiei75 can you please have another look

auvipy (approved):

alirafiei75:
> @alirafiei75 can you please have another look

I was out of town. Looks good to me.

django_celery_beat/schedulers.py
FabianClemenz: This broke my celery beat installation. `timezone.get_current_timezone()` hat the correct timezone "Europe/Berlin" but with the new implementation the code resolves to `UTC` which stopped my tasks from beeing executed after update. Our Server, where our Container running celery with celery beat runs, is using UTC 

This is a breaking change not a little patch update

FabianClemenz: this change caused our environment to stop dispatching tasks https://github.com/celery/django-celery-beat/issues/894 - this should not have been a patch upgrade

FabianClemenz: since you state in the docs that changing the timezone needs manual work https://django-celery-beat.readthedocs.io/en/latest/#important-warning-about-time-zones

auvipy: @alirafiei75 can you also have a look into this please?

alirafiei75: > @alirafiei75 can you also have a look into this please?

I am checking but I do not understand what is the problem at the moment

FabianClemenz: @alirafiei75 the problem is, that on already running instances the scheduling stopped after updating

FabianClemenz: localtime of beat shows time in Europe/Berlin - no matter what timezone the server is in

auvipy:
we got another PR in this front! https://github.com/celery/django-celery-beat/pull/896 please check and let us know!

Issue #809, opened 2024-09-26
DJANGO_CELERY_BEAT_TZ_AWARE conflicts with setting beat_cron_starting_deadline of celery


cy9747:
### Summary:

[beat_cron_starting_deadline](https://docs.celeryq.dev/en/stable/userguide/configuration.html#beat-cron-starting-deadline) is a celery configuration variable.

beat_cron_starting_deadline is used in the [is_due](https://github.com/celery/celery/blob/ad882dc991e46b46c1a195771a897de7127f1e6b/celery/schedules.py#L639C5-L681C28) function of the crontab class, which celery beat's class [TzAwareCrontab](https://github.com/celery/django-celery-beat/blob/8b3992665059172e573ae751851dc2e207761afc/django_celery_beat/tzcrontab.py#L10) inherits from, but overwrites is_due so that this setting does not work.

* Celery Version: 5.4.0
* Celery-Beat Version:  2.7.0

### Exact steps to reproduce the issue:
1. set beat_cron_starting_deadline = 300
2. set DJANGO_CELERY_BEAT_TZ_AWARE = True
3. beat_cron_starting_deadline not work

cy9747:
Is it a good idea to use super?

before

```
def is_due(self, last_run_at):
        last_run_at = last_run_at.astimezone(self.tz)

        rem_delta = self.remaining_estimate(last_run_at)
        rem = max(rem_delta.total_seconds(), 0)
        due = rem == 0
        if due:
            rem_delta = self.remaining_estimate(self.now())
            rem = max(rem_delta.total_seconds(), 0)
        return schedstate(due, rem)
```

after

```
def is_due(self, last_run_at):
    last_run_at = last_run_at.astimezone(self.tz)

    return super().is_due(last_run_at)
```

auvipy:
can you open a PR with failing test case so that we can re produce this?

smiling-watermelon:
@auvipy, as I understand, this should cover what they're talking about.

@cy9747, can you check, and comment/edit if it's not quite right?

auvipy:
thanks for the test, and using super in django model method call is not a bad idea

smiling-watermelon:
BTW, from personal experience the newly updated crontab exclude also _may_ contribute to this problem. (Running code from main branch)

For example, if the task has a schedule like:
[...]

PR #888, opened 2025-05-12
Add test for beat_cron_starting_deadline configuration setting

smiling-watermelon:
Helper PR with two new tests for #809


auvipy (changes requested):
we got new fixes for this with lots of tests. do we need this tests as well?

smiling-watermelon:
Not sure.
Since #809 was closed with a suggestion toedit docs, I believe tests that were added here, to demonstrate the behaviour from afformentioned issue are not necessary.

I'll probably close the PR though, since (at least one of) tests should be rewriten in reverse anyway.

PR #534, opened 2022-04-11
Support for abstract base models improved. (#516)

diegocastrum:
Model classes moved into a `models.py` directory, which exports by default only the default `django-celery-beat` model classes now from `models.generic.py`.

`managers.py` moved into `models.py` directory, because it's something that is needed only by models.

PR #911, opened 2025-06-25
Abstract Django DB Models

blbuford:
Hello,

We encountered an issue in our products where multiple Django services share the same database/schema with Django Celery Beat (DCB) installed. In this setup, Service A attempts to execute Service B’s periodic tasks and vice versa, even though they do not share the same codebase. To address this, we aimed to configure DCB to write to separate tables for each service.

I found an implementation in #534 and have updated it to work with the current mainline DCB. This feature was also requested in #904, and @auvipy mentioned it is on the DCB roadmap. Unit tests pass on my local Python environment, though I have not run the full test matrix.

This change has been very helpful for us, allowing us to do the following:
```python
class ProductAPeriodicTask(AbstractPeriodicTask):

    objects = PeriodicTaskQuerySet.as_manager()

    class Meta:
        db_table = 'django_celery_beat_periodictask_product_a'
        verbose_name = _('ProductA Periodic Task')
        verbose_name_plural = _('ProductA Periodic Tasks')
        abstract = False
```
and add the following to our `settings.py`:
```python
CELERY_BEAT_PERIODICTASK_MODEL = 'core.ProductAPeriodicTask'
```
With this approach, DCB writes to a dedicated table for Product A.

Please let me know if there’s anything I can do to help move this forward or improve the PR for merging.

Thank you!

Best regards,
Brett


blbuford:
Is there any progress or input on this?

PR #931, opened 2025-08-12
refactor: Remove one unnecessary database query

martinlehoux:


martinlehoux:
Hi, sorry if I did not respect some process. I was browsing through this code and saw this inefficiency. Being a refactor, I don't what test I should add, the expected behaviour is exactly the same as before, but removes an extra database query

auvipy:
can you share some screenshots please as starter?

cclauss:
Looking at the original code, I would say that the minimum tests should assert the result where:
1. the happy path is taken (i.e., no exceptions are raised).
2. `cls.DoesNotExist` is raised.
3. `MultipleObjectsReturned` is raised.

Please use those tests to ensure that identical results are returned with the current code and the proposed code.

martinlehoux:
Do you think `class TestDuplicatesMixin` is not sufficient today? I should improve this one test?

Issue #934, opened 2025-08-25
cron_descriptor needs to be locked down as the new release changes the signature.


rh0dium:
cron_descriptor release 2.0.x changes the expected [signatures](https://github.com/Salamek/cron-descriptor/blob/master/cron_descriptor/__init__.py) that celery beat is expecting.  Specifically it looks like
- `MissingFieldException` → ` MissingFieldError`
- `FormatException` → `FormatError`
- `WrongArgumentException` → `WrongArgumentError`

robswc:
Came here to say this, builds have been breaking.  Seems it was released ~30 minutes ago and thought maybe it was just me 😅 

FWIW, I was having no issues with `1.4.5`

sam-14uel:
Same with me

PR #935, opened 2025-08-25
Addresses #934

rh0dium:


requirements/default.txt
```diff
@@ -4,4 +4,4 @@ django-timezone-field>=5.0
 backports.zoneinfo; python_version<"3.9"
 tzdata
 python-crontab>=2.3.4
-cron-descriptor>=1.2.32
+cron-descriptor>=1.2.32,<2.0.0
```
cclauss, line 7:
```suggestion
cron-descriptor>=2.0.0
```
And then modify `django_celery_beat/models.py` to use `FormatError, MissingFieldError, WrongArgumentError` as was done in
* #936

cclauss (approved):
cron-descriptor v2 seems to break too much.

@Salamek

auvipy (approved):

Issue #956, opened 2025-10-02
Crontab tasks permanently excluded due to static time window in optimization


ncimino:
## Description

A critical bug exists in django-celery-beat v2.8.1 where the crontab optimization feature permanently excludes periodic tasks from execution if they fall outside a 2-hour window calculated at scheduler startup. 

Introduced in commit: https://github.com/celery/django-celery-beat/commit/87c059706506994f677e82426c80028ad9894af1

The `_get_crontab_exclude_query()` method creates a static time window that never updates, causing tasks scheduled more than 2 hours from the current server time at startup to never be loaded or executed.

## Environment

- **django-celery-beat version**: 2.8.1
- **Python version**: 3.9+ (uses zoneinfo)
- **Django version**: (any supported version)
- **Celery version**: (compatible with django-celery-beat 2.8.1)

## Steps to Reproduce

1. Start celery beat with DatabaseScheduler at time T (e.g., 10:00 AM)
2. Create a crontab task scheduled to run at T+4 hours (e.g., 2:00 PM) 
3. Wait for the scheduled time
4. Observe that the task never executes
5. Check logs - the task is not loaded into the scheduler's task list

## Expected Behavior

All enabled crontab tasks should be evaluated for execution regardless of when the scheduler started, and tasks should run at their scheduled times.

## Actual Behavior

Tasks scheduled outside the initial 2-hour window (calculated at startup) are permanently excluded from the scheduler and never execute.

## Root Cause

The optimization in `enabled_models_qs()` calls `_get_crontab_exclude_query()` which calculates a static time window:

```python
def _get_crontab_exclude_query(self):
    """
    Build a query to exclude crontab tasks based on their hour value,
    adjusted for timezone differences relative to the server.
    """
    # This calculation happens ONCE at startup and never updates
    server_time = aware_now()  # Static time calculation
    server_hour = server_time.hour
    
    # Creates a static 2-hour window that never changes
    hours_to_include = [
        (server_hour + offset) % 24 for offset in range(-2, 3)
    ]
    hours_to_include += [4]  # celery's default cleanup task
    # ... rest of method
```

The `enabled_models_qs()` method calls this optimization:

```python
def enabled_models_qs(self):
    next_schedule_sync = now() + datetime.timedelta(
        seconds=SCHEDULE_SYNC_MAX_INTERVAL
    )
    exclude_clock_tasks_query = Q(
        clocked__isnull=False,
        clocked__clocked_time__gt=next_schedule_sync
    )

[...]

PR #961, opened 2025-10-08
fix: simplify task fetching by removing crontab exclusion #956

ncimino:
Removed crontab exclusion logic from task fetching. Filters were created to limit window to +/-2 hours for crontab, but tasks are never refetched - must either implement refetching protocol or (as implemented here) remove the filter entirely.

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -284,75 +284,8 @@ class DatabaseScheduler(Scheduler):
             clocked__isnull=False,
             clocked__clocked_time__gt=next_schedule_sync
         )
-
-        exclude_cron_tasks_query = self._get_crontab_exclude_query()
-
-        # Combine the queries for optimal database filtering
-        exclude_query = exclude_clock_tasks_query | exclude_cron_tasks_query
-
         # Fetch only the tasks we need to consider
-        return self.Model.objects.enabled().exclude(exclude_query)
-
-    def _get_crontab_exclude_query(self):
-        """
-        Build a query to exclude crontab tasks based on their hour value,
-        adjusted for timezone differences relative to the server.
-
-        This creates an annotation for each crontab task that represents the
-        server-equivalent hour, then filters on that annotation.
-        """
-        # Get server time based on Django settings
-
-        server_time = aware_now()
-        server_hour = server_time.hour
-
-        # Window of +/- 2 hours around the current hour in server tz.
-        hours_to_include = [
-            (server_hour + offset) % 24 for offset in range(-2, 3)
-        ]
-        hours_to_include += [4]  # celery's default cleanup task
-
-        # Get all tasks with a simple numeric hour value
-        valid_numeric_hours = self._get_valid_hour_formats()
-        numeric_hour_tasks = CrontabSchedule.objects.filter(
-            hour__in=valid_numeric_hours
-        )
-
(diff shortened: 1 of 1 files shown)


ncimino:
> can not verify this as there is no appropriate test cases. and the CI is failing. i think it would be better to implement improved protocol

@auvipy / @alirafiei75 any suggestions on how/when the tasks should be getting refetched/parsed from the DB? The code obviously expected this behavior, but it is not happening.  I'm also curious if there are other tests that we expected to cover this code originally?

kamyanskiy:
> can not verify this as there is no appropriate test cases. and the CI is failing. i think it would be better to implement improved protocol

I think that mechanism with sliding window is not bad, we just need to force this sliding every hour or less. 
In 2.8.1 was added code that seems forces full reload every 5 minutes. So it seems its not a problem anymore https://github.com/celery/django-celery-beat/blob/6427286d59466b4af931739fb47cf61404b14560/django_celery_beat/schedulers.py#L519-L531

auvipy:
but we need a proper solution with appropriate tests

ncimino:
That 5 minute force refresh is not happening for crontab. Some code runs, but the reload never rereads/refreshes from the database.

This code you are referring to as the fix has been there since before v2.8.1 and the issue presents in v2.8.1:
https://github.com/celery/django-celery-beat/blame/v2.8.1/django_celery_beat/schedulers.py#L515-L527

As @auvipy has indicated, I think testing (if representative of client crontab use case) will be quite revealing.

kamyanskiy:
@ncimino @auvipy I would say that on 2.8.1 its not an issue anymore -   just look at my logs - I showed that its work  here https://github.com/celery/django-celery-beat/issues/956#issuecomment-3496313051

so sliding window works (at least with celery 5.5.3), you can see how it forces reload crontab schedules every 300sec and it works as well, my tasks those were planned out of sliding window, finally loaded in memory when window overlaps it,  so I can see crontab tasks in my schedule in (-2 +2h) period. Looks good.

auvipy:
> @ncimino @auvipy I would say that on 2.8.1 its not an issue anymore - just look at my logs - I showed that its work here [#956 (comment)](https://github.com/celery/django-celery-beat/issues/956#issuecomment-3496313051)
> 
> so sliding window works (at least with celery 5.5.3), you can see how it forces reload crontab schedules every 300sec and it works as well, my tasks those were planned out of sliding window, finally loaded in memory when window overlaps it, so I can see crontab tasks in my schedule in (-2 +2h) period. Looks good.

thanks for verifying

cclauss:
% `pip install pre-commit`
% `pre-commit run --all-files`

Or:
% `pipx run pre-commit run --all-files`

Or:
% `uvx pre-commit run --all-files`

auvipy:
> > @ncimino @auvipy I would say that on 2.8.1 its not an issue anymore - just look at my logs - I showed that its work here [#956 (comment)](https://github.com/celery/django-celery-beat/issues/956#issuecomment-3496313051)
> > so sliding window works (at least with celery 5.5.3), you can see how it forces reload crontab schedules every 300sec and it works as well, my tasks those were planned out of sliding window, finally loaded in memory when window overlaps it, so I can see crontab tasks in my schedule in (-2 +2h) period. Looks good.
> 
> thanks for verifying

closing in favor of this

PR #964, opened 2025-10-24
Calculate time_since_last_sync in UTC time.

Hatell:
Fixes issue when DST is changed between calculation.


auvipy:
please also check the failing tests due to this change

cclauss:
https://docs.python.org/3/library/datetime.html#datetime.UTC was added in Python 3.11.

Hatell:
I make test to this. When drafting test I noticed that `int(time.time())` might be better in this use case.

Issue #755, opened 2024-06-12
Remove Strict Upper Version Limit for Django


Mogost:
Hi Maintainers,

I am writing to request the removal of the strict upper version limit on Django in the `django-celery-beat` library, with the added bonus of ensuring compatibility with Django 5.1.

**Background:**

Currently, the library enforces a strict upper limit on the supported versions of Django. This has previously caused issues for users wanting to upgrade to newer Django versions, as noted in the following issues:
- Issue [#680](https://github.com/celery/django-celery-beat/issues/680): Users requested compatibility with Django 5.0, highlighting the problem of strict version constraints.
- Issue [#698](https://github.com/celery/django-celery-beat/issues/698): Similar concerns were raised about the need for Django 5.0 support.

**Arguments for Removing the Upper Version Limit:**

1. **Increased Flexibility:** Removing the upper version limit will provide users with the flexibility to upgrade to newer versions of Django as soon as they are released, without waiting for an official update from `django-celery-beat`.

2. **Accelerated Compatibility Testing:** By allowing the installation of the library with newer Django versions, the community can quickly test and report any issues, facilitating faster updates and fixes from the maintainers.

3. **Clear Compatibility Declaration:** While removing the upper limit doesn't implicitly declare support for all future Django versions, it allows developers to experiment and provide feedback, thereby contributing to a more robust testing process.

4. **Community Feedback:** There has been consistent feedback from the community about the inconveniences caused by the strict version constraints. Addressing this would significantly improve the developer experience for a wide range of users.

**Proposal:**

1. Remove the upper version limit for Django in the `setup.py` and `tox.ini` files.
2. Ensure compatibility with Django 5.1 by running the test suite and making any necessary adjustments.

Thank you for considering this request. Your efforts in maintaining this essential library are greatly appreciated by the community.

auvipy:
is django 5.1 released yet? or any alpha yet?

Mogost:
> is django 5.1 released yet? or any alpha yet?

yep. https://pypi.org/project/Django/5.1a1/ alpha

auvipy:
thanks, that is enough!

auvipy:
the upper limit is relaxed already https://github.com/celery/django-celery-beat/blob/main/requirements/test-django.txt

Mogost:
[...]

PR #971, opened 2025-11-18
Remove upper bound on Django version

realsuayip:
Why?

Adding upper bound on Django makes it hard to test the library in new Django versions (even if there is no official support).

Currently i'm trying to use Django 6.0 alpha but django-celery-beat won't let me. This is not productive because:

a) users won't be able test the library against new Django versions, making them unable to report potential bugs beforehand
b) in case a new django version is found compatible, users still need to update django-celery-beat (which they might not want to do)


cclauss:
A long discussion of why we have been setting an upper limit when we have lots of user issues:
* #755

realsuayip:
> A long discussion of why we have been setting an upper limit when we have lots of user issues:
> 
> * #755

Thanks for the reference. I have now read the majority of the discussion and still not convinced that strict upper pinning is required. And the community seems to support this too.

Of course, I'm not asking people to create a library that potentially supports all future Django versions. This is a separate concern aka 'confirmed support'

When users upgrade their Django and break django-celery-beat, they can create issues to ask for 'official support' that's fine. But withholding them from upgrading is not useful at all (see my points above)

What if somebody breaks their setup without realizing / haphazardly upgrading?

The answer is that's their responsibility; this is why lockfiles exist. People should not upgrade Django without checking confirmed support from Django packages.

cclauss:
Help is needed on the 100+ open issues.
* https://github.com/celery/django-celery-beat/issues

requirements/runtime.txt
```diff
@@ -1 +1 @@
-Django>=2.2,<6.0
+Django>=2.2
```
auvipy:
```suggestion
Django>=2.2,<6.1
```

auvipy (changes requested):
this is the saner option

auvipy:
> But withholding them from upgrading is not useful at all (see my points above)

no one is withholding anyone. anyone can maintain their separate forks to meet and fit their needs

realsuayip:
> anyone can maintain their separate forks to meet and fit their needs

Yes, but how is that useful for `django-celery-beat` though? Do people really have to maintain a fork just to test `django-celery-beat` with newer Django versions?

I'm not asking for a super-specific feature that merits it own fork. And I'm not asking for unrealistic support/maintenance either.

auvipy:
> > anyone can maintain their separate forks to meet and fit their needs
> 
> Yes, but how is that useful for `django-celery-beat` though? Do people really have to maintain a fork just to test `django-celery-beat` with newer Django versions?
> 
> I'm not asking for a super-specific feature that merits it own fork. And I'm not asking for unrealistic support/maintenance either.

every project has its own maintenance support policy and time constraint. if that does‘t suite some ones need and they are in a hurray, they can maintain their own fork until a new release come. it is kind of the norm.

PR #996, opened 2026-02-03
feat: add hebrew locale

ShlomoCode:


Nusnus (commented):
LGTM - not sure if RTL will be applied correctly though.

@ShlomoCode did you check if the mixed English-Hebrew strings appear correctly?

PR #1045, opened 2026-05-28
fix: cap is_due delay at start_time for one-off crontab tasks (#1044)

### LINKED ISSUE #1044: Regression since 2.8.0 (#844): crontab PeriodicTask with future start_time wakes far in the future via due_start_time(), firing up to SCHEDULE_SYNC_MAX_INTERVAL late
## Summary

Since **2.8.0** (PR #844), a `PeriodicTask` backed by a `CrontabSchedule` with a future `start_time` is woken up by `DatabaseScheduler` far later than `start_time`. For a one-shot task whose crontab pins a single calendar date, the wake-up is scheduled **years** out, so the task only runs when the periodic forced full sync (`SCHEDULE_SYNC_MAX_INTERVAL`, 300s) reloads the schedule and finds it overdue — in practice firing **up to ~5 minutes late**. This worked correctly in **2.7.0**.

## Affected versions

- Broken: 2.8.0, 2.8.1, **2.9.0** (latest)
- Working: 2.7.0
- celery 5.x, `DatabaseScheduler`, `USE_TZ=True`, `TIME_ZONE="UTC"` — a non-UTC timezone is **not** required; all-UTC reproduces it.

## Root cause

`ModelEntry.is_due`, in the `now < start_time` branch:

**2.7.0 (correct):**
```python
delay = math.ceil((self.model.start_time - now).total_seconds())
return schedules.schedstate(False, delay)
```

**2.8.0–2.9.0 (regressed, via #844):**
```python
start_time = self.model.due_start_time(current_tz)
delay = math.ceil((start_time - now).total_seconds())
return schedules.schedstate(False, delay)
```

For a crontab-backed task, `PeriodicTask.due_start_time()` → `CrontabSchedule.due_start_time()`:
```python
def due_start_time(self, initial_start_time, tz):
    start_time = initial_start_time.astimezone(tz)
    start, ends_in, now = self.schedule.remaining_delta(start_time)
    return start + ends_in   # next crontab match STRICTLY after start_time
```

`remaining_delta(start_time)` returns the next instant matching the crontab *after* `start_time`. When the crontab pins `day_of_month` + `month_of_year` + `day_of_week` to one calendar date (the natural way to express a one-shot run), the next re-match is years away (e.g. 2026-05-27 → 2037-05-27). So `is_due` returns a delay of ~11 years; the scheduler heap never wakes at `start_time`. Only the forced full sync (`SCHEDULE_SYNC_MAX_INTERVAL = 300`) reloads the entry and runs it once already overdue. (Same #844 change as the sibling report #875.)

## Minimal reproduction

```python
from datetime import timedelta
from django.utils import timezone
from django_celery_beat.models import PeriodicTask, CrontabSchedule

run_at = timezone.now() + timedelta(minutes=2)
cron, _ = CrontabSchedule.objects.get_or_create(
    minute=str(run_at.minute), hour=str(run_at.hour),
    day_of_month=str(run_at.day), month_of_year=str(run_at.month),
    day_of_week=str(run_at.isoweekday() % 7),
)
task = PeriodicTask.objects.create(
    name="one-shot repro", task="celery.ping",
    crontab=cron, start_time=run_at, one_off=True,
    expires=run_at + timedelta(minutes=15),
)

print("start_time     :", task.start_time)
print("due_start_time :", task.due_start_time(timezone.get_current_timezone()))
# -> due_start_time prints a datetime YEARS in the future, not run_at
```

Run `celery beat` with `DatabaseScheduler`: the task fires ~`SCHEDULE_SYNC_MAX_INTERVAL` after `start_time`, not at `start_time`.

## Expected

The task becomes due at (≈) `start_time`, as in 2.7.0.

## Suggested fix

In `ModelEntry.is_due`, when `now < start_time`, the retry delay should not push the wake-up past `start_time` itself:
```python
start = self.model.due_start_time(current_tz)
start = min(start, self.model.start_time)   # never wake later than start_time
```
(PR #844 targeted a different start_time calculation; the crontab `due_start_time` path over-corrects for one-shot tasks.)

## Workaround

Use `ClockedSchedule` for one-shot absolute-time tasks — `PeriodicTask.due_start_time()` returns `start_time` directly when there is no crontab, so it is unaffected.

yanlong-pan:
For a one-off PeriodicTask whose CrontabSchedule pins day_of_month + month_of_year (and often day_of_week) to specific values, CrontabSchedule.due_start_time() returns the *next* matching calendar date -- typically one year in the future. ModelEntry.is_due() then returned this years-long delay, causing the task to fire ~5 minutes late (when the forced sync interval expires) instead of at start_time.

Cap the start_time used for delay computation by min()-ing with the user-specified model.start_time when model.one_off is True. Recurring crontabs are unaffected -- day_of_month='*' makes the next match at most 24h away, and the post-#844 wake-at-next-match behavior is desirable for them.

Fixes #1044

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -124,6 +124,15 @@ class ModelEntry(ScheduleEntry):
                 # send a delay to retry on start_time
                 current_tz = now.tzinfo
                 start_time = self.model.due_start_time(current_tz)
+                if self.model.one_off:
+                    # For a one-off PeriodicTask whose CrontabSchedule pins
+                    # day_of_month + month_of_year + day_of_week to fixed
+                    # values, CrontabSchedule.due_start_time() returns the
+                    # *next* matching calendar date, which is years in the
+                    # future. Without this cap, the task is delayed ~5
+                    # minutes (until the forced sync interval expires)
+                    # instead of firing at start_time. See issue #1044.
+                    start_time = min(start_time, self.model.start_time)
                 time_remaining = start_time - now
                 delay = math.ceil(time_remaining.total_seconds())
 
--- a/t/unit/test_schedulers.py
+++ b/t/unit/test_schedulers.py
@@ -1033,6 +1033,76 @@ class test_DatabaseScheduler(SchedulerCase):
         assert not is_due
         assert next_check == pytest.approx(expected_delay, abs=60)
 
+    def test_oneoff_crontab_fixed_date_does_not_delay_past_start_time(
+            self, app):
+        """Regression test for issue #1044.
+
+        A one-off PeriodicTask with a CrontabSchedule pinned to a specific
+        calendar date (day_of_month + month_of_year fixed) must wake up at
+        start_time, not at the next matching crontab date -- which can be
+        years in the future.
+
+        Before PR #844 (released in 2.8.0), is_due returned a delay of
+        ``(start_time - now)``. After #844, it returns
+        ``(crontab.due_start_time - now)``, which for fixed-date crontabs
+        computes the *next* matching date -- typically one year out for a
+        day_of_month + month_of_year pin -- causing tasks to fire ~5
+        minutes late (when the forced sync interval expires) instead of
+        at start_time.
+        """
+        now = app.now()
+
+        delay_minutes = 2
+        test_start_time = now + timedelta(minutes=delay_minutes)
+
+        task = self.create_model_crontab(
+            crontab(
+                minute=f'{test_start_time.minute}',
+                hour=f'{test_start_time.hour}',
+                day_of_month=f'{test_start_time.day}',
+                month_of_year=f'{test_start_time.month}',
+            ),
+            start_time=test_start_time,
+            one_off=True,
+        )
+
+        entry = EntryTrackSave(task, app=app)
+        is_due, next_check = entry.is_due()
+
+        expected_delay = delay_minutes * 60
+
+        assert not is_due
+        assert next_check == pytest.approx(expected_delay, abs=60)
+
+    def test_recurring_crontab_with_start_time_unaffected_by_oneoff_cap(
+            self, app):
+        """The fix for #1044 only caps one_off tasks; recurring crontabs
+        are unaffected by the cap and continue to use the post-#844
+        behavior (wake up at the next crontab match after start_time).
+        """
+        now = app.now()
+
+        delay_minutes = 2
+        test_start_time = now + timedelta(minutes=delay_minutes)
+        crontab_time = test_start_time + timedelta(minutes=delay_minutes)
+
+        task = self.create_model_crontab(
+            crontab(minute=f'{crontab_time.minute}'),
+            start_time=test_start_time,
+            one_off=False,
+        )
+
+        entry = EntryTrackSave(task, app=app)
+        is_due, next_check = entry.is_due()
+
+        # Same expectation as
+        # test_crontab_with_start_time_between_now_and_crontab
+        # (4 minutes = start_time delay + crontab delay).
+        expected_delay = 2 * delay_minutes * 60
+
+        assert not is_due
+        assert next_check == pytest.approx(expected_delay, abs=60)
+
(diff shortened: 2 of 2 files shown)


auvipy (commented):
so we are OK with partial changes for the regression?

JinRiYao2001:
Author of #844 here, just wanted to share some notes from digging into this. The core question I keep coming back to is whether `start_time >= cron match time` should be considered a valid configuration — most of the rest follows from that.

**The repro**

```python
run_at = timezone.now() + timedelta(minutes=2)  # e.g. 05:36:53.747
cron = CrontabSchedule(
    minute=str(run_at.minute),   # "36"
    hour=str(run_at.hour),       # "5"
    day_of_month=str(run_at.day),
    month_of_year=str(run_at.month),
    day_of_week=str(run_at.isoweekday() % 7),
)
task = PeriodicTask(crontab=cron, start_time=run_at, one_off=True)
```

The cron pattern matches at minute boundary (e.g. `05:36:00`). `start_time` is set to `run_at`, which falls in the same minute as the cron tick (`05:36:53` here, but the seconds don't matter — the boundary is the minute).

Whenever `start_time` lands in the same minute as the cron tick (or later) for a fixed-date crontab, `CrontabSchedule.due_start_time()` returns the *next* matching calendar date — typically years out. I verified this on the celery side: `crontab.remaining_delta()` uses strict `<` on `last_run_at.minute < max(self.minute)`, so once `start_time.minute == cron.minute`, that occurrence is treated as already passed regardless of seconds. Only `start_time` strictly before the cron-tick minute (e.g. `05:35:59`) is handled correctly.

In 2.7.0 this didn't surface because the delay was just `start_time - now` with no crontab lookup at all.

**What #1045 actually does**

The patch does `min(due_start_time, model.start_time)` for one-off tasks. Since `due_start_time` is by definition `>= model.start_time`, the `min()` is unconditionally `model.start_time`. So for one-off crontab tasks, the patch reverts the delay calculation back to pre-#844's `start_time - now`. Recurring crontab keeps #844's behavior. In that sense it's closer to a scoped revert of #844 on the one-off path than a scoped fix.

**A question worth raising**

The original issue #843 that #844 was addressing wasn't about the task failing to fire — it was about heap-top blocking. In celery's scheduler `tick()`:

```python
event = H[0]                      # peek at top, don't pop
is_due, _ = self.is_due(entry)
if is_due:
    heappop(H)                    # only pop when due
    ...
# if not due, the top stays, and entries below it are never evaluated this tick
```

Pre-#844, a crontab task with future `start_time` had heap priority equal to `start_time` (absolute). If that was the smallest priority in the heap, the entry sat at the top with `is_due=False` for the entire wait, and other entries weren't evaluated until it cleared. #844 pushed the priority forward to the next cron tick so the task moves down the heap and other entries can be processed normally during the wait.

With #1045, one-off crontab tasks with future `start_time` go back to having heap priority = `start_time`. Would this reintroduce the same heap-top blocking shape from #843 for one-off tasks with future `start_time`? Worth a closer look before merging.

**Two ways to look at the policy call**

- If `start_time` falling in the same minute as the cron tick (or later) is treated as an invalid config, this isn't really a bug — users should set `start_time` strictly before the cron tick, or use `ClockedSchedule` for absolute single fires (which the issue itself already lists as the workaround, and which `PeriodicTask` already validates as `one_off=True`). Since celery crontab is minute-resolution (no seconds field), the cron tick is always at `:00` of some minute, so `start_time < cron match time` naturally implies `start_time` lives in an earlier minute and the same-minute case is avoided by construction.
- The 2.7.0 backward-compat angle is worth weighing against the actual impact surface. The justification really only holds if a meaningful number of users are relying on this shape in production — but from what's visible here, it looks like a single reporter's case rather than a broader pattern (the issue has been silent from other users, and `ClockedSchedule` already covers the one-off absolute-time use case cleanly). If you do want to land a fix anyway, it'd be worth the PR author taking another pass on (a) whether the heap-top blocking from #843 comes back on the one-off path, and (b) the same minute-boundary misalignment exists for non-one-off fixed-date crontabs too, which the current patch doesn't cover.

PR #1060, opened 2026-07-01
pre-commit: Add ruff rules for unsafe naive datetimes

### LINKED ISSUE #1059: DatabaseScheduler.schedule().current_time is a naive datetime
Is using a [`naive datetime`](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects) here a problem?

```python
class DatabaseScheduler(Scheduler):
    [ ... ]
```
https://github.com/celery/django-celery-beat/blob/6594458128ee08c98e6dada23d5fd2eaa726f972/django_celery_beat/schedulers.py#L515-L518
% `ruff check --select=DTZ005 ` # https://docs.astral.sh/ruff/rules/call-datetime-now-without-tzinfo
```
DTZ005 `datetime.datetime.now()` called without a `tz` argument
```
* #879
* #1060

@alirafiei75

---
~Also, the variable `update` defined on line 517 will ALWAYS be True (line 522 or 527), so the code on lines 532 to 542 will NEVER be executed.~

cclauss:
Fixes: #1059

Lint rules to ban the usage of unsafe [`naive datetimes`](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects).
* https://docs.astral.sh/ruff/rules/#flake8-datetimez-dtz

% `ruff check --select=DTZ --statistics`
```
7	DTZ001	call-datetime-without-tzinfo
7	DTZ005	call-datetime-now-without-tzinfo
5	DTZ003	call-datetime-utcnow
Found 19 errors.
```
Related to:
* #1059

### DIFF
--- a/django_celery_beat/schedulers.py
+++ b/django_celery_beat/schedulers.py
@@ -159,7 +159,7 @@ class ModelEntry(ScheduleEntry):
         else:
             # this ends up getting passed to maybe_make_aware, which expects
             # all naive datetime objects to be in utc time.
-            now = datetime.datetime.utcnow()
+            now = datetime.datetime.utcnow()  # noqa: DTZ003 We want a naive datetime
         return now
 
     def __next__(self):
@@ -391,7 +391,7 @@ class DatabaseScheduler(Scheduler):
         target_tz = ZoneInfo(timezone_name)
 
         # Use a fixed point in time for the calculation to avoid DST issues
-        fixed_dt = datetime.datetime(2023, 1, 1, 12, 0, 0)
+        fixed_dt = datetime.datetime(2023, 1, 1, 12, 0, 0, tzinfo=server_tz)
 
         # Calculate the offset
         dt1 = fixed_dt.replace(tzinfo=server_tz)
@@ -515,7 +515,7 @@ class DatabaseScheduler(Scheduler):
     @property
     def schedule(self):
         initial = update = False
-        current_time = datetime.datetime.now()
+        current_time = aware_now()
 
         if self._initial_read:
             debug('DatabaseScheduler: initial read')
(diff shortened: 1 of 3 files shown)


django_celery_beat/schedulers.py
```diff
@@ -391,7 +391,7 @@ def _get_timezone_offset(self, timezone_name):
         target_tz = ZoneInfo(timezone_name)
 
         # Use a fixed point in time for the calculation to avoid DST issues
-        fixed_dt = datetime.datetime(2023, 1, 1, 12, 0, 0)
+        fixed_dt = datetime.datetime(2023, 1, 1, 12, 0, 0, tzinfo=server_time.tzinfo)
```
alirafiei75: I think it is better to use server_tz which has been defined a couple lines above.

alirafiei75 (approved):

django_celery_beat/schedulers.py
```diff
@@ -515,7 +515,7 @@ def _refresh_schedule(self):
     @property
     def schedule(self):
         initial = update = False
-        current_time = datetime.datetime.now()
+        current_time = aware_now()
```
auvipy, line 518: I think this change and the one above demands unit tests for verification. @alirafiei75 what do you think?

alirafiei75: I do not think that is critical because both of the changes do not make any logical difference and there are already unit tests on the function. but it wouldn't hurt.

django_celery_beat/schedulers.py
```diff
@@ -416,7 +416,7 @@ def _get_timezone_offset(self, timezone_name):
         target_tz = ZoneInfo(timezone_name)
 
         # Use a fixed point in time for the calculation to avoid DST issues
-        fixed_dt = datetime.datetime(2023, 1, 1, 12, 0, 0)
+        fixed_dt = datetime.datetime(2023, 1, 1, 12, 0, 0, tzinfo=server_tz)
```
auvipy, line 419: dont we need to test this with unit tests?

cclauss: No.  That was the other one that was reverted.  As @alirafiei75 said, they were both no-ops.

PR #1061, opened 2026-07-03
Fix lock contention in PeriodicTasks change detection

alirafiei75:
## Summary

This PR refactors how django-celery-beat detects schedule changes for Celery Beat. The previous design updated a **single singleton row** (`PeriodicTasks`, `ident=1`) on **every** insert, update, and delete via Django signals. Under high concurrency—especially with `ATOMIC_REQUESTS=True` and long transactions—workers contended for an exclusive lock on that row, causing PostgreSQL `lock_timeout` errors ([issue #1000](https://github.com/celery/django-celery-beat/issues/1000)).

The new design **stops writing the shared singleton on normal saves**. Beat detects inserts/edits by reading timestamps that already exist on the task/schedule rows. Deletes and bulk updates still bump the singleton, but only after `commit`.

Under `ATOMIC_REQUESTS` / `atomic()`, a real schedule edit also gets a **short post-commit restamp** of that same task/schedule row (`date_changed` / `updated_at`). That is an extra **per-row** write, not a return of the old `ident=1` hotspot.

## Root cause

`PeriodicTasks.update_changed()` used `update_or_create()`, which in modern Django acquires `SELECT FOR UPDATE` on the singleton row. Even replacing that with a plain `UPDATE` does not fully solve the problem: **any row update holds a lock until `COMMIT`**. When the signal runs inside a long `atomic()` block (e.g. `ATOMIC_REQUESTS=True`), all concurrent workers serialize on the same row.

### Pull-based detection for writes (no singleton write)

Beat reads the latest change timestamp from existing / new `auto_now` fields:

- `PeriodicTask.date_changed` (already existed; now indexed)
- `updated_at` on `IntervalSchedule`, `CrontabSchedule`, `SolarSchedule`, `ClockedSchedule` (new, indexed)

`PeriodicTasks.last_change()` returns:

```text
max(
  MAX(PeriodicTask.date_changed),
  MAX(IntervalSchedule.updated_at),
  MAX(CrontabSchedule.updated_at),
  MAX(SolarSchedule.updated_at),
  MAX(ClockedSchedule.updated_at),
  PeriodicTasks.last_update,  # singleton marker
)
```

Inserts, updates, and in-place schedule edits advance one of the `MAX(...)` values automatically—**no singleton write**.

### Commit-time restamp inside `atomic()`

`auto_now` stamps **save time**, not **commit time**. Overlapping transactions can therefore commit out of order relative to those stamps and hide a change behind Beat's watermark.

So for saves already inside an atomic block:

- after `COMMIT`, restamp that row's `date_changed` / `updated_at`
- autocommit saves stay a single write
- Beat housekeeping (`last_run_at` / `total_run_count`) is skipped so Beat does not reload itself
- restamp / marker callback errors are logged and swallowed so a successful commit cannot become a 500

If the process dies between `COMMIT` and the restamp, that change may stay below the watermark until Beat's periodic full sync (`SCHEDULE_SYNC_MAX_INTERVAL`, ~5 minutes).

### Singleton row repurposed for out-of-band changes only

It is bumped only when timestamps cannot help:

- **Deletes** (row gone; `MAX` may drop or stay unchanged)
- **Admin bulk actions** (`queryset.update()` bypasses `auto_now`)

Writes are deferred with **`transaction.on_commit()`** so the lock is not held for the duration of the outer transaction.

### Partial `save(update_fields=...)`

Django skips `auto_now` when the tracking field is omitted from `update_fields`. Real edits now include `date_changed` / `updated_at` so Beat still notices them.

### Signal wiring

| Event | Before | After |
|-------|--------|-------|
| `PeriodicTask` save | `pre_save` + explicit `save()` hook | *(none — `date_changed` + optional commit restamp)* |
| `PeriodicTask` delete | `pre_delete` | `post_delete` |
| Schedule model save | `post_save` | *(none — `updated_at` + optional commit restamp)* |
| Schedule model delete | `pre_delete` / `post_delete` | `post_delete` marker bump |

### Critical: beat housekeeping must not bump `date_changed`

`ModelEntry.save()` uses `update_fields=['last_run_at', 'total_run_count']`. A full `save()` would bump `date_changed` on every task run, and `MAX(date_changed)` would make Beat reload the schedule in an **infinite loop**. An inline comment documents this constraint; a regression test guards it.

## Migration

`0021_clockedschedule_updated_at_and_more.py` (depends on upstream `0020` enabled-index migration):

- Adds indexed `updated_at` to the four schedule models
- Adds index on `PeriodicTask.date_changed`

On large tables this can be expensive (table rebuilds on older MySQL; blocking index builds on PostgreSQL unless you use `CREATE INDEX CONCURRENTLY`). Plan a maintenance window or apply equivalent index DDL out-of-band if needed.

## Alternatives considered

| Approach | Why not chosen |
|----------|----------------|
| Plain `UPDATE` without `SELECT FOR UPDATE` | Still holds row lock until `COMMIT`; does not fix `ATOMIC_REQUESTS` contention |
| `transaction.on_commit()` bump of the singleton on every save | Avoids long lock hold, but brings write traffic back to the shared hotspot |
| Full-table diff / fingerprint each Beat tick | Correct under reordered commits, but `O(table)` and unsafe at multi-million scale |
| INSERT new row per change + cleanup job | Eliminates contention but adds write amplification, table growth, cleanup ops, and migration complexity |
| In-memory global flag | Beat and workers are separate processes; no shared memory |
| Postgres `LISTEN/NOTIFY` | Fast but DB-specific; breaks the library’s DB-agnostic design |

## Trade-offs

- **Pros:** No singleton write/lock on normal saves; deletes still prompt; in-place schedule edits detected; partial `update_fields` saves stay visible; out-of-order commits covered by post-commit restamp; minimal schema change; 5-minute forced sync remains as safety net
- **Cons:**
  - `last_change()` runs several indexed `MAX()` queries per Beat tick (cheap vs. former lock storms)
  - under `ATOMIC_REQUESTS`, each real edit does an extra short per-row restamp after commit
  - `update_changed()` is async via `on_commit` (a read of `last_change()` in the same open transaction will not see the bump yet)
  - admin bulk updates still require explicit `update_changed()`
  - if restamp never runs (crash / swallowed DB error), detection may wait for the periodic full sync

## Breaking note

`PeriodicTasks.last_update` is an internal marker and is **no longer advanced on every save**. Use `PeriodicTasks.last_change()` (what Beat itself uses). See Changelog for the full upgrade note.

After bulk `queryset.update()`, call `PeriodicTasks.update_changed()` only. `PeriodicTasks.changed()` is the delete signal handler, not the bulk-update API.

## Test plan

- [x] Unit: save advances `last_change()` via `date_changed` (no marker bump)
- [x] Unit: delete advances `last_change()` via marker (`on_commit` capture)
- [x] Unit: in-place schedule edit advances `last_change()` for all four schedule types
- [x] Unit: beat housekeeping save does **not** bump `date_changed` or trigger reload
- [x] Unit: bulk `queryset.update()` + `update_changed()` advances marker
- [x] Unit: deleting row with max `date_changed` remains detectable (marker dominates)
- [x] Unit: late-committing older save stamp is still detected (task + schedule)
- [x] Unit: `save(update_fields=...)` omitting tracking fields is still detected
- [x] Unit: autocommit save does not register a commit restamp
- [x] Unit: failed restamp / marker `on_commit` callbacks are logged, not raised

## References

- https://github.com/celery/django-celery-beat/issues/1000


django_celery_beat/models.py
```diff
@@ -437,14 +463,40 @@ def changed(cls, instance, **kwargs):
...
+        for model, field in (
+            (PeriodicTask, 'date_changed'),
+            (IntervalSchedule, 'updated_at'),
+            (CrontabSchedule, 'updated_at'),
+            (SolarSchedule, 'updated_at'),
+            (ClockedSchedule, 'updated_at'),
+        ):
+            val = model.objects.aggregate(m=Max(field))['m']
+            if val:
+                stamps.append(val)
```
cclauss: Call `stamps.extend()` once instead of repeated calls to `stamps.append()`.  See: [`ruff rule FURB113`](https://docs.astral.sh/ruff/rules/repeated-append)
```suggestion
        stamps.extend(
            val
            for model, field in (
                (PeriodicTask, "date_changed"),
                (IntervalSchedule, "updated_at"),
                (CrontabSchedule, "updated_at"),
                (SolarSchedule, "updated_at"),
                (ClockedSchedule, "updated_at"),
            )
            if (val := model.objects.aggregate(m=Max(field))["m"])
        )
```

django_celery_beat/models.py
```diff
@@ -437,14 +463,40 @@ def changed(cls, instance, **kwargs):
...
+
+        transaction.on_commit(_bump)
 
     @classmethod
     def last_change(cls):
+        stamps = []
         try:
-            return cls.objects.get(ident=1).last_update
+            marker = cls.objects.get(ident=1).last_change_marker
+            if marker:
```
cclauss:
```suggestion
            if marker := cls.objects.get(ident=1).last_change_marker:
```

django_celery_beat/models.py
```diff
@@ -415,16 +440,17 @@ def due_start_time(self, initial_start_time, tz):
...
+    This stores a single row with ``ident=1``. ``last_change_marker`` is
+    bumped for changes not captured by ``auto_now`` timestamps on
+    :class:`~.PeriodicTask` or schedule models (deletions and admin bulk
+    ``queryset.update()``). Inserts and in-place edits are detected by
+    reading ``MAX(date_changed)`` / ``MAX(updated_at)`` instead.
     """
 
     ident = models.SmallIntegerField(default=1, primary_key=True, unique=True)
-    last_update = models.DateTimeField(null=False)
+    last_change_marker = models.DateTimeField(null=False)
```
auvipy: i think we can keep the attribute name as is, as there is no actual change here

django_celery_beat/signals.py
```diff
@@ -9,37 +9,22 @@ def signals_connect():
...
     )
 
-    signals.post_save.connect(
-        PeriodicTasks.update_changed, sender=IntervalSchedule
-    )
     signals.pre_delete.connect(
         PeriodicTasks.update_changed, sender=IntervalSchedule
     )
 
-    signals.post_save.connect(
```
auvipy, line 26: changes here are breaking change. are they strictly necessary? can we figure out any backward compatible way?

django_celery_beat/models.py
```diff
@@ -114,6 +115,12 @@ class SolarSchedule(models.Model):
         help_text=_('Run the task when the event happens at this longitude'),
         validators=[MinValueValidator(-180), MaxValueValidator(180)],
     )
+    updated_at = models.DateTimeField(
+        auto_now=True,
+        db_index=True,
```
auvipy: for db_index should we consider index in meta option?

auvipy (changes requested):

django_celery_beat/models.py
alirafiei75: I just wanted this to be clear that the column has different usage now. I will revert it.

django_celery_beat/signals.py
alirafiei75: this is the main reason for PR. if I do not delete the post save signals, the lock duration stays the same under heavy load. with this change the beat just reads the save changes from another source.

django_celery_beat/models.py
alirafiei75: yes meta is a better practice. I will change this and check the copilot comments too and will get back to you. thanks.

django_celery_beat/signals.py
```diff
@@ -9,37 +9,22 @@ def signals_connect():
                          IntervalSchedule, PeriodicTask, PeriodicTasks,
                          SolarSchedule)
 
-    signals.pre_save.connect(
-        PeriodicTasks.changed, sender=PeriodicTask
-    )
     signals.pre_delete.connect(
         PeriodicTasks.changed, sender=PeriodicTask
     )
```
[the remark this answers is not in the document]
alirafiei75: the change applied and I agree.

django_celery_beat/migrations/0020_rename_last_update_periodictasks_last_change_marker_and_more.py
```diff
@@ -0,0 +1,55 @@
...
+            old_name="last_update",
+            new_name="last_change_marker",
+        ),
+        migrations.AddField(
+            model_name="clockedschedule",
+            name="updated_at",
+            field=models.DateTimeField(
+                auto_now=True, db_index=True, null=True, verbose_name="Last Modified"
+            ),
+        ),
```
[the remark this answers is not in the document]
alirafiei75: this is not necessary. the description was wrong and I will change it

django_celery_beat/migrations/0020_clockedschedule_updated_at_and_more.py
```diff
@@ -0,0 +1,70 @@
+# Generated by Django 5.0.1 on 2026-07-05 07:08
```
auvipy: would you mind removing it and generate a new one with the newer changes? also it would be nice to use atleast django 5.2.x LTS

auvipy (changes requested):

django_celery_beat/migrations/0020_clockedschedule_updated_at_and_more.py
alirafiei75: this one has the new changes but I will now change it and generate it with lts

rodrigondec (commented):
Thank your for the effort @alirafiei75!


## Overview

This PR replaces the singleton marker row (`PeriodicTasks`, `ident=1`, updated via `update_or_create`/lock) with pull-based detection: `last_change()` is now a `MAX()` over `date_changed` (PeriodicTask) and `updated_at` (new, on the 4 schedule models), plus the singleton marker, which is only touched for cases `auto_now` can't cover (deletes, bulk `queryset.update()`). Writes to the marker are deferred via `transaction.on_commit()`. This fixes the root cause of the contention (#1000): it removes the write from the save hot path entirely, rather than just swapping `SELECT FOR UPDATE` for a plain `UPDATE` (which would still hold the row lock until `COMMIT`).

I like this solution — I agree it's better than the fix I proposed in the original issue. This PR's pull-based approach eliminates the write from the hot path altogether.

## Code quality & style

Well structured, comments land in the right places (the warning in `ModelEntry.save()` about not adding `date_changed` to `update_fields` is essential — without it, a future "fix" reintroduces the infinite reload loop). Migration is auto-generated and clean, indexes match the fields used in the `MAX()` queries.

## Findings

1. **Silent compatibility break** — `PeriodicTasks.last_update` is a public model field. Any external code (dashboards, monitoring scripts, third-party libs) reading that field directly instead of calling `last_change()` will now see a **stale** value, with no error and no warning.

   After this PR, `last_update` only advances in 3 cases, all through `update_changed()`:
   - **Delete** — `post_delete` on `PeriodicTask` or any of the 4 schedule models fires `changed()`/`update_changed()` directly.
   - **Admin bulk update** — `enable_tasks`, `disable_tasks`, `toggle_tasks` (`admin.py`) use `queryset.update()`, which bypasses `auto_now`; that's why they call `update_changed()` manually right after.
   - **First creation of the singleton row** — if `ident=1` doesn't exist yet, `update_changed()` creates it via a fallback (`.create()` with `IntegrityError` handled for the concurrent-creation race).

   A normal save of `PeriodicTask` or any schedule model **no longer** touches `last_update` — what advances instead is `date_changed`/`updated_at` (auto_now) on those tables themselves, and only `last_change()` (the `MAX()` aggregating the marker plus those columns) reflects the real state. Before this PR, `last_update` was bumped on every save/delete, so this is a new, non-obvious behavior for anyone going by the field name alone. Worth an explicit breaking-change note in the changelog/docs — not just the class docstring — making clear that `last_update` in isolation is no longer the source of truth, only `last_change()` is.

2. **Per-tick cost on Beat (suggestion, non-blocking)** — `last_change()` now runs up to 6 queries (1 `get` on the marker + 5 `aggregate`/`MAX`, one per model) every tick, versus 1 query before. Indexes make each `MAX()` cheap in isolation, but that's 6 network/DB round-trips instead of 1 — adds up with a short tick interval (seconds) and many Beat instances running. The PR already acknowledges this trade-off.

   This could be collapsed into a single query via `UNION ALL` over 6 subselects, avoiding the round-trips — but that steps outside plain Django ORM (would need `RawSQL`/`.raw()` or a manual cursor), which counts against readability and ORM-only familiarity for maintainers. Worth considering as an optimization if the overhead proves relevant in practice, not as a blocker — the current version is simpler to read and maintain, and the trade-off (6 lightweight indexed queries vs. 1 less-readable query) may not be worth it without measuring the actual cost first.

3. **Post-migration `NULL` window (informational only)** — the `AddField` for `updated_at` is `auto_now=True, null=True` with no `default`, so existing schedule rows will have `updated_at=NULL` until their next save. This doesn't affect the correctness of `MAX()` — SQL/Django aggregates ignore `NULL`, so the computed value stays correct across the populated rows. A row that's never re-saved simply doesn't contribute to `last_change()` until its first save after this deploy (expected, since nothing actually changed on it). The one event that would matter in that window — deleting a never-touched row — is already covered separately by the explicit marker bump via `post_delete`, regardless of whether `updated_at` was `NULL`. No correctness risk, just a behavioral detail worth knowing about post-deploy.

4. **`no_changes` flag is now dead for the `PeriodicTask` save path** — `_disable()` and the `one_off` branch in `is_due()` still set `self.model.no_changes` before `save()`, but since `PeriodicTasks.changed(self)` was removed from `PeriodicTask.save()` entirely (not just gated, removed), that flag no longer suppresses or triggers anything on that path — it only survives because `_refresh_schedule()` still copies it around in memory. A comment at those two call sites would prevent future confusion about what the flag actually does now.

5. **Minor** — `admin.py` calls `update_changed()` unconditionally in all 3 actions (`enable_tasks`, `disable_tasks`, `toggle_tasks`), even when `queryset.update()` returns 0 affected rows:

   ```python
   def enable_tasks(self, request, queryset):
       rows_updated = queryset.update(enabled=True)
       # queryset.update() bypasses auto_now on date_changed; bump marker.
       PeriodicTasks.update_changed()
       self.message_user(request, ...)
   ```

   `rows_updated` is computed but never used as a guard. In practice, `rows_updated == 0` only happens in a rare edge case (rows selected in the admin got deleted by another process between selection and clicking the action, or an empty queryset via direct URL manipulation) — but when it does, the code still fires `update_changed()` unnecessarily: it schedules an `on_commit` callback, issues an unneeded UPDATE/CREATE against the `PeriodicTasks` table, and forces Beat to reload the whole schedule on the next tick despite nothing having actually changed.

   Trivial fix, same pattern in all 3 actions:
   ```python
   if rows_updated:
       PeriodicTasks.update_changed()
   ```

   Non-blocking — just a wasted write/reload in a rare edge case, doesn't cause inconsistency.

## Test coverage

Good — covers save without marker bump, delete via marker, in-place edits for all 4 schedule types, the housekeeping-save guard, bulk update, and the `IntegrityError` fallback. Missing: a test that explicitly documents that `PeriodicTasks.last_update` (the raw field) goes stale after a normal save — locking in this behavior as intentional and preventing a future regression that "fixes" it back.

## Recommendation

The design is correct and addresses the root cause, not just the symptom. Before approving I'd want: (1) an explicit breaking-change note in the PR/changelog for finding 1, (2) confirmation of the per-tick query cost for finding 2. `auvipy` already requested changes — worth checking whether they overlap with these points.

alirafiei75:
> Thank your for the effort @alirafiei75!

thanks for your detailed review.
I will check your points thoroughly.

alirafiei75:
Thanks for the thorough review @rodrigondec 

Finding 1 (compatibility break): Added an explicit breaking-change note in the Changelog and a test (test_normal_save_does_not_bump_last_update_field) that locks in this behavior.

Finding 2 (per-tick queries): Agreed. keeping the ORM-only version for readability.

Finding 3 (NULL window): Acknowledged, has no risk.

Finding 4 (no_changes flag): Added clarifying comments at _disable(), the one-off branch, and __next__().

Finding 5 (admin guard): Wrapped all three actions with if rows_updated:.

auvipy (changes requested):
I am reluctant on introducing yet another breaking change. any way to keep backward compatible?

alirafiei75:
> I am reluctant on introducing yet another breaking change. any way to keep backward compatible?

@auvipy I hear the concern, I think the only way to keep backward compatibility and also keep the pr changes is to introduce a new config so people can change it in their settings and have these new optimizations, but I think a config adds complexity for an internal detail that shouldn't be user-facing. last_update was never a documented API — last_change() is the documented interface (and what Beat itself uses). The field still exists and still updates on deletes/bulk actions; it just isn't the sole source of truth anymore.
Even with on_commit, writing to the same row on every save across many concurrent workers is still significant I/O pressure on that single row — that's the core problem this PR solves.
Happy to strengthen the changelog/upgrade note, but I'd rather not add a config toggle for an internal mechanism. What do you think?

rodrigondec (approved):

auvipy:
> > I am reluctant on introducing yet another breaking change. any way to keep backward compatible?
> 
> @auvipy I hear the concern, I think the only way to keep backward compatibility and also keep the pr changes is to introduce a new config so people can change it in their settings and have these new optimizations, but I think a config adds complexity for an internal detail that shouldn't be user-facing. last_update was never a documented API — last_change() is the documented interface (and what Beat itself uses). The field still exists and still updates on deletes/bulk actions; it just isn't the sole source of truth anymore. Even with on_commit, writing to the same row on every save across many concurrent workers is still significant I/O pressure on that single row — that's the core problem this PR solves. Happy to strengthen the changelog/upgrade note, but I'd rather not add a config toggle for an internal mechanism. What do you think?

yeah that would be acceptable

alirafiei75:
> > > I am reluctant on introducing yet another breaking change. any way to keep backward compatible?
> > 
> > 
> > @auvipy I hear the concern, I think the only way to keep backward compatibility and also keep the pr changes is to introduce a new config so people can change it in their settings and have these new optimizations, but I think a config adds complexity for an internal detail that shouldn't be user-facing. last_update was never a documented API — last_change() is the documented interface (and what Beat itself uses). The field still exists and still updates on deletes/bulk actions; it just isn't the sole source of truth anymore. Even with on_commit, writing to the same row on every save across many concurrent workers is still significant I/O pressure on that single row — that's the core problem this PR solves. Happy to strengthen the changelog/upgrade note, but I'd rather not add a config toggle for an internal mechanism. What do you think?
> 
> yeah that would be acceptable

Great! I'll update the PR to polish the upgrade notes/changelog to make the behavior change crystal clear

auvipy:
please fix the newly raised merge conflicts

alirafiei75:
> please fix the newly raised merge conflicts

@auvipy I did resolve the conflicts.

django_celery_beat/models.py
```diff
@@ -437,14 +471,40 @@ def changed(cls, instance, **kwargs):
...
+        stamps.extend(
+            val
+            for model, field in (
+                (PeriodicTask, 'date_changed'),
+                (IntervalSchedule, 'updated_at'),
+                (CrontabSchedule, 'updated_at'),
+                (SolarSchedule, 'updated_at'),
+                (ClockedSchedule, 'updated_at'),
+            )
+            if (val := model.objects.aggregate(m=Max(field))['m'])
```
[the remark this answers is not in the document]
alirafiei75: Good catch. I would restamp date_changed/updated_at on that row after commit, only inside atomic(). Also regression test added.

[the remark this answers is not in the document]
alirafiei75: save() now adds the tracking field so auto_now actually runs. Housekeeping (last_run_at/total_run_count) is skipped. Also added tests.

django_celery_beat/migrations/0020_clockedschedule_updated_at_and_more.py
```diff
@@ -0,0 +1,70 @@
+# Generated by Django 5.2.15 on 2026-07-05 12:14
```
auvipy: I think we have to remove and regenerate it after pulling from the main branch

auvipy (changes requested):
also we need to fix the newly introduced merge conflicts

django_celery_beat/models.py
```diff
@@ -738,6 +738,10 @@ class Meta:
         verbose_name = _('periodic task')
         verbose_name_plural = _('periodic tasks')
         indexes = [
+            models.Index(
+                fields=['enabled'],
+                name='beat_periodic_enabled_idx',
+            ),
             models.Index(fields=['date_changed']),
```
auvipy, line 783: please cross check this part

auvipy:
@mehrazrumman it would be great if you can try and review this carefully

django_celery_beat/models.py
alirafiei75: Yeah it was the part that was added by this pr and the previous merged pr. It is ok I think. Is anything wrong in your opinion @auvipy ?

MehrazRumman:
OK @auvipy  !

MehrazRumman (changes requested):
## Review

I checked out the branch and ran the unit suite locally: **173 passed**. The one failure (`test_modeladmin_PeriodicTaskAdmin::test_run_task`) reproduces identically on `main` and only needs a live broker, so it is unrelated to this PR.

This is a well-reasoned fix for #1000. Moving from "write the singleton on every save" to "read `MAX()` of columns that already exist" removes the write from the hot path entirely, rather than just trading `SELECT FOR UPDATE` for a plain `UPDATE`, which would still hold the row lock until `COMMIT`. The alternatives table in the description is genuinely useful, and the earlier review rounds have already closed the big ones: the partial-save gap, the out-of-order-commit watermark hole, the duplicate bump from the old `PeriodicTask.delete()` override (now removed), the `pre_delete` → `post_delete` ordering, and the admin `if rows_updated:` guards. The inline warning on `ModelEntry.save()` about never adding `date_changed` to `update_fields` is exactly the right comment in exactly the right place.

Below is the one thing I think still needs a change, plus some smaller notes.

### Should be fixed before merge

**Deferred callbacks have no error handling, and a failure there turns a committed write into a 500.**

Neither `_touch()` in `_restamp_tracking_field_on_commit()` nor `_bump()` in `update_changed()` is wrapped in `try/except`. Django runs `on_commit` callbacks after `COMMIT` but still inside the caller's stack, and an exception there propagates to the caller while the data stays committed. I confirmed this on the branch:

```
--- deferred callback raises after a successful commit ---
RESULT: exception PROPAGATED to caller -> Exception: transient DB error in deferred bump
data committed anyway? name = p2
```

Under `ATOMIC_REQUESTS=True`, which is the exact deployment this PR targets, a transient DB hiccup during the restamp or the marker bump would return a 500 for a request whose data was already written. `_bump()` handles `IntegrityError` for the creation race but nothing else.

Suggestion: wrap both callback bodies in `try/except Exception` and log a warning. A missed timestamp bump is self-healing, since the `max_interval` forced sync picks the change up, so failing loudly here costs far more than it saves.

### Worth a note, not blocking

1. **"No extra writes on the hot path" is not quite accurate under `ATOMIC_REQUESTS`.** The restamp only fires inside an atomic block, which under `ATOMIC_REQUESTS=True` means *every* non-housekeeping save. So each save now costs two writes instead of one. They are per-row writes with no shared-row contention, so the lock problem really is solved, but the write count goes up rather than down for the target audience. Worth restating honestly in the description and Changelog so operators size the change correctly.

2. **Residual miss window.** If the process dies between `COMMIT` and the restamp `UPDATE`, that change stays stamped at save time, below Beat's watermark, and is only picked up by the `max_interval` sync. That is a dramatic narrowing of the original hole rather than a closure, which seems like the right trade, but it deserves a sentence in the docstring so the next reader does not assume the guarantee is absolute.

3. **Migration cost on large tables.** `0021` adds four columns and five indexes in one migration. On MySQL before 8.0 `ADD COLUMN` rebuilds the table, and a plain `CREATE INDEX` on PostgreSQL blocks writes for the duration. Deployments big enough to hit #1000 are exactly the ones with a large `periodictask` table. A short upgrade note suggesting `CREATE INDEX CONCURRENTLY` or a maintenance window would help.

4. **Positional `update_fields`.** `_ensure_tracking_in_update_fields()` reads `kwargs.get('update_fields')`, so a legacy positional call such as `obj.save(False, False, None, ['args'])` silently skips change detection. Passing positionally is deprecated in modern Django, so this is an edge case, but it is a silent failure rather than a loud one.

5. **`update_changed()` is now asynchronous.** It is documented as the bulk-update API, and after this change a caller inside `atomic()` who calls it and then reads `last_change()` in the same transaction sees a stale value. Probably harmless in practice, worth one line in the docstring.

### On the backward-compatibility question

For what it is worth, I agree with the resolution already reached with @auvipy. `last_update` was never documented, `last_change()` is what Beat itself uses, and the field still exists and still advances on deletes and bulk actions. The Changelog entry spells out the exact behaviour change and names the migration path, and `test_normal_save_does_not_bump_last_update_field` locks it in. A settings toggle for an internal mechanism would mean maintaining two change-detection paths forever, which seems worse than a clearly documented note.

Nice work on this one. Address the callback error handling and I think it is ready.

alirafiei75:
Thanks for the careful review @MehrazRumman.

Agreed on the should-fix: both _touch() and _bump() now catch exceptions, log a warning with traceback, and leave the committed write alone. Missed bumps still get picked up by the periodic full sync.

Also took the soft notes: Changelog/README now say clearly that under ATOMIC_REQUESTS you get an extra per-row restamp (not a return of the singleton hotspot), docstring covers the die-between-commit-and-restamp window and that update_changed() only lands after commit, plus a short migration cost note for large tables.

MehrazRumman (approved):

django_celery_beat/models.py
```diff
@@ -89,7 +93,88 @@ def crontab_schedule_celery_timezone():
...
+            logger.warning(
+                'Failed to restamp %s.%s for pk=%s after commit: %r',
+                model_cls.__name__, field_name, pk, exc,
+                exc_info=True,
+            )
+
+    transaction.on_commit(_touch, using=using)
+
+
+class TimestampedScheduleMixin(models.Model):
```
auvipy, line 160: with introduction of this, the existing systems in production might face some glitch.... what do you think?

alirafiei75: The mixin only adds updated_at and a short post-commit restamp inside atomic(). It does not change when tasks run, so I do not know what you mean. If you have a specific glitch in mind, I can address that.

auvipy:
@cclauss please give a look into it and provide review. after that we can decide how to proceed with this

PR #1072, opened 2026-07-31
Optimize enabled schedule query

### LINKED ISSUE #1071: DatabaseScheduler clocked-task filter forces a LEFT OUTER JOIN and full PeriodicTask scan
### Summary

`DatabaseScheduler.enabled_models_qs()` currently excludes distant one-off tasks with a nullable relation lookup:

```python
Q(
    clocked__isnull=False,
    clocked__clocked_time__gt=next_schedule_sync,
)
```

Django implements this as a `LEFT OUTER JOIN` from `PeriodicTask` to `ClockedSchedule`. The scheduler does not select any column from that relation, and the base query has no index for its mandatory `enabled=True` predicate. Installations retaining many disabled historical tasks can therefore scan the full `PeriodicTask` table whenever beat reloads its schedule.

This behavior was introduced by #835 and is present in 2.8.x, 2.9.0, and current `main`.

### Why the outer join is unnecessary

The existing rule excludes a task only when its related clocked schedule is after the next-sync horizon. It can be expressed directly as an anti-join:

```python
future_clocked_tasks = ClockedSchedule.objects.filter(
    pk=OuterRef("clocked_id"),
    clocked_time__gt=next_schedule_sync,
)

PeriodicTask.objects.enabled().filter(~Exists(future_clocked_tasks))
```

For a non-clocked task the subquery has no row, so `NOT EXISTS` includes it. Clocked tasks at or before the cutoff remain included, while later tasks are excluded. This preserves the current semantics and keeps `enabled_models_qs()` lazy and filterable without the nullable outer join.

A positive formulation such as `clocked_id IS NULL OR EXISTS(due clocked schedule)` is logically equivalent, but it is not a good replacement here: on PostgreSQL it can inflate the correlated subplan cost enough to trigger JIT when enabled tasks are common. The direct `NOT EXISTS(future clocked schedule)` form avoids that regression.

### Reproduction / query-plan evidence

PostgreSQL 17 benchmarks covered two synthetic workload shapes.

With 500,000 mostly-disabled historical tasks and 1,200 enabled tasks:

| Query | Enabled index | Execution time |
| --- | --- | ---: |
| Current `LEFT OUTER JOIN` | no | 27.772 ms |
| Positive `OR EXISTS` | no | 84.955 ms |
| Current `LEFT OUTER JOIN` | yes | 0.844 ms |
| Proposed `NOT EXISTS` | yes | 0.869 ms |

With 100,000 tasks, 90% enabled, and 1% clocked:

| Query | Execution time |
| --- | ---: |
| Current `LEFT OUTER JOIN` | 11.784 ms |
| Positive `OR EXISTS` | 96.932 ms |
| Proposed `NOT EXISTS` | 10.911 ms |

The first workload shows that removing the join alone is not sufficient: an enabled-task index is what prevents the large base-table scan. The second shows why the anti-join is preferable to the positive correlated form. Exact timings vary, but `NOT EXISTS` remains stable across both selectivity profiles.

A plain `PeriodicTask(enabled)` index is portable across PostgreSQL, MySQL/MariaDB, and SQLite, while allowing each optimizer to ignore it when the Boolean field is not selective. In a 90%-enabled SQLite benchmark, the normal scan took 41.8 ms and forcing equality/index use took 48.5 ms, so the implementation should not force SQLite to select the index.

### Proposed change

1. Replace the clocked-task outer-join exclusion with `NOT EXISTS` for a future related clocked schedule.
2. Add `Index(fields=["enabled"], name="beat_periodic_enabled_idx")`.
3. Preserve the existing `filter(enabled=True)` behavior and schedule prefetches.
4. Add regression coverage for:
   - enabled non-clocked and eligible clocked tasks;
   - disabled and distant-future clocked tasks;
   - exact cutoff inclusivity;
   - lazy queryset composition and schedule prefetches;
   - generated SQL containing `NOT EXISTS` and no clocked `LEFT OUTER JOIN`;
   - the supporting index remaining in model/migration state.

A tested fix is available in #1072.

AryanHamedani:
Fixes #1071

## Summary

- replace the clocked-task `LEFT OUTER JOIN` with an exact `NOT EXISTS` anti-join
- add a portable index for the scheduler's mandatory `enabled` filter
- preserve the lazy/filterable `enabled_models_qs()` extension point and its schedule prefetches
- cover candidate semantics, the exact cutoff, SQL shape, index state, and prefetch behavior

## Rationale

The scheduler never selects fields from `ClockedSchedule`; it only excludes a task when its related clocked schedule is beyond the next-sync horizon. The nullable relation join is therefore unnecessary. Expressing the existing rule directly as `NOT EXISTS` keeps null/non-clocked tasks eligible without either a nullable outer join or a positive `clocked_id IS NULL OR EXISTS(...)` predicate.

The anti-join shape matters. During review, the positive `OR EXISTS` form was rejected because PostgreSQL assigned the correlated subplan a high enough cost to trigger JIT when enabled tasks were common. `NOT EXISTS(future clocked schedule)` produced a low-cost anti-join instead.

The query rewrite alone cannot avoid scanning a large `PeriodicTask` table when only a small fraction of rows are enabled, so this also adds a simple index on `enabled`. A plain index is portable across the supported databases and lets each optimizer decide whether the field is selective enough to use it. The implementation deliberately keeps the existing `filter(enabled=True)` spelling; forcing SQLite to select a low-cardinality Boolean index was slower when most tasks were enabled.

## Query-plan evidence

Synthetic PostgreSQL 17 benchmarks covered both relevant selectivity profiles:

| Dataset | Current outer join | Positive `OR EXISTS` | Proposed `NOT EXISTS` |
| --- | ---: | ---: | ---: |
| 500k mostly-disabled historical tasks, enabled index on proposed query | 27.772 ms (no new index) | 84.955 ms (no new index) | 0.869 ms |
| 100k tasks, 90% enabled, 1% clocked | 11.784 ms | 96.932 ms | 10.911 ms |

On the mostly-disabled dataset, the current query with the new index took 0.844 ms, confirming that the index addresses the base-table scan while `NOT EXISTS` removes the unnecessary join without the positive form's JIT regression. Timings are workload-dependent; the important result is the stable plan shape across both selectivity regimes.

On SQLite with 90% enabled rows, leaving Django's normal Boolean predicate produced a 41.8 ms scan while forcing equality/index use took 48.5 ms. The PR therefore does not prescribe SQLite's plan.

## Compatibility

The candidate set is unchanged:

- enabled non-clocked tasks remain included
- past, near-future, and exact-cutoff clocked tasks remain included
- clocked tasks after the cutoff remain excluded
- disabled tasks remain excluded
- the existing crontab exclusion is unchanged

`Exists` and `OuterRef` are available throughout the project's Django 3.2+ support range. The returned value remains a lazy, composable queryset and retains the four existing schedule prefetches.

## Verification

- complete unit suite: 147 passed (144 in the full run plus 3 broker-dependent admin tests with `CELERY_BROKER_URL=memory://`)
- Django 3.2 focused semantics, cutoff, and prefetch tests: 3 passed
- `python manage.py makemigrations --check --dry-run`: no changes detected
- Flake8 on all affected Python files: passed
- `git diff --check`: passed
- PostgreSQL 17 plans benchmarked under both mostly-disabled and mostly-enabled task distributions
- migration model state, forward application, and reverse application verified during review


AryanHamedani:
Hi @auvipy, just checking in on this one. I fixed the only thing the Copilot review flagged back on Aug 1 (the test cutoff now uses a 1-second delta instead of 1 microsecond), and CI has been green across the whole matrix since. It still merges cleanly against main, and nothing else has claimed the 0020 migration slot yet.

Is there anything you'd like me to change before this can move forward? The benchmarks behind the change are over in #1071 if useful.

auvipy:
@alirafiei75 can you cross check this?

alirafiei75:
Looks good to me @auvipy. NOT EXISTS instead of the outer join is the right shape, the enabled index makes sense for large mostly-disabled tables, and the tests cover the important bits.

alirafiei75 (approved):

auvipy (approved):

auvipy (approved):

